PR #114, opened 2022-08-18
[xml-ingress] ChildIdentifiersTable, ChildCharacteristicsTable, CINdetailsTable

tab1tha:
Continues from https://github.com/data-to-insight/CIN-validator/pull/113

**Purpose:**
Create a table each for CINdetails, ChildCharacteristics and ChildIdentifiers to contain all the tag-value pairs that do not have subtags under them.


tab1tha:
Todo

- Generate multiple rows when a child has more than one of any of the subsections. Right now in the case of multiple CINdetails blocks, causes the previous table values to be overwritten by the next.
- Generate multiple rows of each DataFrame when more than one child is present. Right now, the next child overwrites the values of the previous child.

tab1tha:
Todo

Disabilities should not be present in the ChildCharacteristics table. That's a bug.

PR #117, opened 2022-09-08
Rule template: Demonstrate on Rule 8500

tab1tha:
This pull request uses the implementation of rule 8500 to check that we have a solid template for creating CIN rules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+__pycache__/
+*.pyc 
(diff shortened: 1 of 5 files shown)


tab1tha:
Thought: How does location linking work if one of the conditions is that a value is missing. That is, the error is raised  because a child is absent from a table. How will that be flagged in the table?.

kws:
Who don't you do a rule, and then we can look at what a good result API
would look like?

PR #120, opened 2022-09-23
Rule 1520 created, checks for duplicate UPNs

phoebefryer:


tab1tha:
Hi @phoebefryer thank you for this.

Could you do the beginner rules first?. This was marked as `advanced` in the excel sheet because it requires that all the children are compared to each other. 

The beginner rules are set up to compare values in one child. So although this seems very similar to the rule that you just completed, that small detail makes a significant difference. 
No worries at all, we'll get to these types of rules in no time !.

tab1tha:
Hi @phoebefryer , I copied your code over to this branch https://github.com/data-to-insight/CIN-validator/pull/258 because the `phoebefryer:rule_1520` could not be found.

PR #122, opened 2022-09-23
Create rule_8840.py

willLPD2I:
I made this rule in a new branch from my other rule 8840 branch because I want the other one open for writing documentation still!

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8840.py
+++ b/cin_validator/rules/cin2022_23/rule_8840.py
@@ -1,0 +1,79 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import rule_definition, CINTable, RuleContext
+from cin_validator.rule_engine import IssueLocator
+from cin_validator.test_engine import run_rule
+
+# Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
+# Replace ChildIdentifiers with the table name, and LAChildID with the column name you want.
+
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8840.py
```diff
@@ -0,0 +1,79 @@
...
+    # replace the table and column name as done earlier. 
+    # The last numbers represent the index values where you expect the sample data to fail the validation check.
+    assert issues == [
+        IssueLocator(CINTable.ChildProtectionPlans, CPPstartDate, 0),
+        IssueLocator(CINTable.ChildProtectionPlans, CPPstartDate, 1),
+        IssueLocator(CINTable.ChildProtectionPlans, CPPstartDate, 3),
+        IssueLocator(CINTable.ChildProtectionPlans, CPPendDate, 0),
+        IssueLocator(CINTable.ChildProtectionPlans, CPPendDate, 1),
+        IssueLocator(CINTable.ChildProtectionPlans, CPPendDate, 3),
+    ]
```
tab1tha, line 73: Hi Will, 

This is a good line of thought.
Hoewever, for rules that flag more than one location per instance, we'll need to find a way to show the linked locations. I haven't built out that part yet. That's why we are starting with the beginner rules which all flag one location per instance since only one column is involved.

When we've figured out how best to show linking between locations, the structure here will change.
I should have noticed this earlier when you mentioned that you were working on a rule that involved two columns.

Can we put this on hold for a  while?.

PR #124, opened 2022-09-28
Added Rule 1510

PhilDeakin:
Added Rule 1510... 
(Note I couldn't work out how to run the rule to check test_validate is doing what I expect it to)
Also this push, pull, commit, sync, git stuff is very confusing.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_1510.py
+++ b/cin_validator/rules/cin2022_23/rule_1510.py
@@ -55,7 +55,7 @@ def validate(
     df2['LAST_C'] = df2['LAST_C'].apply(lambda x: int(x) if str(x).isdigit() else pd.NA)
     df2 = df2[df2['LAST_C'].notna()]
     for i in range(1,13):
-        df2['C' + str(i)] = (df['UPN'].str[i].astype(int)**(i+1)) % 23      #Largest possible value before mod 23 is applied is 9^13 can Python handle that as an int??
+        df2['C' + str(i)] = (df['UPN'].str[i].astype(int) * (i+1)) % 23
     df2['SUMMED'] = df2[['C1','C2','C3','C4','C5','C6','C7','C8','C9','C10','C11','C12']].sum(axis=1) % 23
 
     check_map = enumerate(list('ABCDEFGHJKLMNPQRTUVWX'))
--- a/cin_validator/rules/cin2022_23/rule_1510.py
+++ b/cin_validator/rules/cin2022_23/rule_1510.py
@@ -64,7 +64,7 @@ def validate(
 
     df2 = df2[df2['CHECK_CHAR'].astype(str) != df2['FIRST_CHAR'].astype(str)]
 
-    failing_indices = df2['index'].index
+    failing_indices = df2.set_index('index').index
 
     rule_context.push_issue(
         table=ChildIdentifiers, field=UPN, row=failing_indices
--- a/cin_validator/rules/cin2022_23/rule_1510.py
+++ b/cin_validator/rules/cin2022_23/rule_1510.py
@@ -29,7 +29,7 @@ def validate(
     # Replace ChildIdentifiers with the name of the table you need.
     df = data_container[ChildIdentifiers]
 
-    # implement rule logic as descriped by the Github issue. Put the description as a comment above the implementation as shown.
+    # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
     """
     <UPN> (N00001) if present must contain the correct check letter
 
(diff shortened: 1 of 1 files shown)


tab1tha:
Hi Phil, I can relate to that. Learning git was an interesting process for me too. We can go over it again during the drop-in tomorrow.

To run the coded rules, move to parent folder of the cin_validator folder in the terminal and do
`python -m cin_validator test`

We can go through this tomorrow also.

tab1tha:
I'm on mobile now so my laptop isn't closeby. 
Thank you for implementing this.
I'll give it a look.

PhilDeakin:
Passes pytest test... which is nice.

cin_validator/rules/cin2022_23/rule_1510.py
```diff
@@ -0,0 +1,96 @@
...
+    rule_context.push_issue(
+        table=ChildIdentifiers, field=UPN, row=failing_indices
+    )
+
+
+def test_validate():
+    # Create some sample data such that some values pass the validation and some fail.
+
+    child_identifiers = pd.DataFrame([['W351201321005'], [pd.NA], ['W35320152101A']], columns=[UPN])
+
```
tab1tha, line 78: As far as testing parts of your functionality to see whether they work as you expect, this is the piece of code that I used to verify some parts of your logic.

```
import pandas as pd
df2 = pd.DataFrame([['W351201321005'], ['A123456789123'], ['W353201521010']], columns=["UPN"])
for i in range(1,13):
    df2['C' + str(i)] = (df2['UPN'].str[i].astype(int)* (i+1))
df2['SUMMED'] = df2[['C1','C2','C3','C4','C5','C6','C7','C8','C9','C10','C11','C12']].sum(axis=1) % 23
print(df2)
```

Of course, we can be more methodical about this. I'll look into that.

cin_validator/rules/cin2022_23/rule_1510.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    df.reset_index(inplace=True)
+    df2 = df[['index','UPN']]
+    df2 = df2[(df2['UPN'].str.len() == 13) & df2['UPN'].notna()]
+    df2['FIRST_CHAR'] = df2['UPN'].str[:1]
+    df2['LAST_C'] = df2['UPN'].str[-12:]
+    df2['LAST_C'] = df2['LAST_C'].apply(lambda x: int(x) if str(x).isdigit() else pd.NA)
+    df2 = df2[df2['LAST_C'].notna()]
+    for i in range(1,13):
+        df2['C' + str(i)] = (df2['UPN'].str[i].astype(int) * (i+1)) % 23
```
tab1tha, line 58: You've divided by 23 twice: here and in the next line.
Check the rule definition one more time and verify if you had to do it here @PhilDeakin ?

cin_validator/rules/cin2022_23/rule_1510.py
```diff
@@ -0,0 +1,96 @@
...
+    df2['SUMMED'] = df2[['C1','C2','C3','C4','C5','C6','C7','C8','C9','C10','C11','C12']].sum(axis=1) % 23
+
+    check_map = enumerate(list('ABCDEFGHJKLMNPQRTUVWX'))
+    check_map = dict((i, j) for i, j in check_map)
+    df2['CHECK_CHAR'] = df2['SUMMED'].map(check_map)
```
tab1tha, line 63: If you were wondering how to create a UPN that'll pass the rule, this is what I did:

When I ran this line, I printed out df2 so that I could see the CHECK_CHAR column. Then I adjusted one of the values in my fake data so that the first character will be equal to CHECK_CHAR for that UPN.

cin_validator/rules/cin2022_23/rule_1510.py
```diff
@@ -0,0 +1,96 @@
...
+    df2['FIRST_CHAR'] = df2['UPN'].str[:1]
+    df2['LAST_C'] = df2['UPN'].str[-12:]
+    df2['LAST_C'] = df2['LAST_C'].apply(lambda x: int(x) if str(x).isdigit() else pd.NA)
+    df2 = df2[df2['LAST_C'].notna()]
+    for i in range(1,13):
+        df2['C' + str(i)] = (df2['UPN'].str[i].astype(int) * (i+1)) % 23
+    df2['SUMMED'] = df2[['C1','C2','C3','C4','C5','C6','C7','C8','C9','C10','C11','C12']].sum(axis=1) % 23
+
+    check_map = enumerate(list('ABCDEFGHJKLMNPQRTUVWX'))
+    check_map = dict((i, j) for i, j in check_map)
```
tab1tha, line 62: This is good work !.

cin_validator/rules/cin2022_23/rule_1510.py
```diff
@@ -0,0 +1,96 @@
...
+    df2['CHECK_CHAR'] = df2['SUMMED'].map(check_map)
+
+    df2 = df2[df2['CHECK_CHAR'].astype(str) != df2['FIRST_CHAR'].astype(str)]
+
+    failing_indices = df2['index'].to_list()
```
tab1tha, line 67: @PhilDeakin passing in the index to rule_context.push_issue() is sufficient. You don't need to convert it into a list anymore as we did in the 903 project.

tab1tha (changes requested):
Thank you so much. 
This was a pleasant review task for me. Most of the text is notes on how else you could have done stuff. 
I have tagged you on the lines that need a few changes. 
Welcome back to the team @PhilDeakin  !

cin_validator/rules/cin2022_23/rule_1510.py
PhilDeakin: So just...
failing_indices = df2['index'].index
? @tab1tha

cin_validator/rules/cin2022_23/rule_1510.py
PhilDeakin: It doesn't make any difference as it will result in the same number.

cin_validator/rules/cin2022_23/rule_1510.py
PhilDeakin: Thanks.

PhilDeakin:
Changes made..

cin_validator/rules/cin2022_23/rule_1510.py
tab1tha: @PhilDeakin 
Apologies for not replying sooner. 

We need the index object that's why we aren't converting to a list. Also, since the saved the original index in the index column (when you did reset_index) to prevent it from getting corrupted, we need to set that original index back in its place by doing df.set_index("index"). 

That is,
After you do `df2 = df2[df2['CHECK_CHAR'].astype(str) != df2['FIRST_CHAR'].astype(str)]`
to get a dataframe containing all the rows where the UPN values are wrong, 

do `failing_indices = df2.set_index('index').index` instead, 

then  `rule_context.push_issue(table=ChildIdentifiers, field=UPN, row=failing_indices)` as you did.

cin_validator/rules/cin2022_23/rule_1510.py
```diff
@@ -0,0 +1,100 @@
...
+    #These should fail
+    'R247962919251', #5 Wrong initial char
+    'X428558133462'  #6 Wrong initial char
+    ]})
+
+    result = run_rule(validate, {ChildIdentifiers: child_identifiers})
+
+    issues = list(result.issues)
+
+    #assert len(issues) == 2
```
tab1tha, line 92: Yes, you are correct. Two error positions will be flagged (index 5 and 6 as you pointed out below). 
You can uncomment this line.

cin_validator/rules/cin2022_23/rule_1510.py
tab1tha: That's interesting. Is there a reason why you chose to do it twice then?
I'm just curious.

tab1tha:
I've resolved the rest of the conversation and left only the updated comments based on your changes. 
We're almost there :)

@PhilDeakin . I guess we can sync on whether you prefer to do these types of finishing touches or if you'll prefer that I do them.

cin_validator/rules/cin2022_23/rule_1510.py
PhilDeakin: The numbers in the C1-12 columns can themselves end up being bigger than 23, so if you apply the modulo there too, it will reduce the size of the integers that have to be summed.... but I doubt in the grand scheme of things it would make any difference timewise as none of the numbers are massive and addition and modulus aren't expensive functions.

PhilDeakin:
One last question (for today anyway)... this obvs works
python -m cin_validator test

But how do I get the pytest result in verbose format. (as it's more useful)

cin_validator/rules/cin2022_23/rule_1510.py
PhilDeakin: Amended it to something neater that avoids the creation of the 12 C columns.

tab1tha (approved):

PR #126, opened 2022-10-03
Rule1530

PhilDeakin:
Added 1530


PhilDeakin:
Seems to have bought the code for 1510 with it... which is annoying.

tab1tha:
I fixed it by running `git merge main` on this branch after I had merged in the changes on Rule1510.
Then I pushed so that Github knows that those changes exist on main already so there's no need to put them in the diff.

PR #143, opened 2022-10-13
Added8545Q

PhilDeakin:
Added8545Q


tab1tha:
Update. 
Hi Phil, Thank you.  This works.
The reason I haven't merged it yet is because I'm testing out a way of creating the period of census centrally so that we do not have to pull in the header table whenever the period of census is needed.

cin_validator/rules/cin2022_23/rule_8545Q.py
```diff
@@ -0,0 +1,106 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import (
+    CINTable,
+    IssueLocator,
+    RuleContext,
+    RuleType,
```
tab1tha, line 9: ruletypes are considered as `error` by default. 
When a rule is a `query` (it's code ends in Q/the github issue is marked as query), then the RuleType can be specified by importing RuleType as shown and assigning the the Query value to the rule context as shown in line 24.

This might not be  in the documentation because few rules are queries so I typically add this part when I noticed that a query rule has been implemented.

cin_validator/rules/cin2022_23/rule_8545Q.py
```diff
@@ -0,0 +1,106 @@
...
+    # Create a key column to join
+    # df["Key"] = 0
+    """df_ref = df_ref[[ReferenceDate]]
+    df_ref["YearEnd"] = pd.to_datetime(df_ref["ReferenceDate"], format=r"%Y-%m-%d", errors="coerce") #equals 31/03/2023  (for example, year can be different)
+    df_ref["YearStart"] = df_ref["YearEnd"] - pd.DateOffset(years=1) + pd.DateOffset(days=1)         #equals 01/04/2022  (for example, year can be different)
+    #Create a key column to join
+    df_ref["Key"] = 0"""
+    # df = df.merge(df_ref, on="Key")
+
+    collection_year = df_ref[Year]
```
tab1tha, line 52: The way you used `ReferenceDate` here was correct also.
I changed it to use Year because that's what the `make_census_period` function expects.

cin_validator/rules/cin2022_23/rule_8545Q.py
```diff
@@ -0,0 +1,106 @@
...
+    failing_indices = df.index
+
+    rule_context.push_issue(
+        table=ChildIdentifiers, field=PersonDeathDate, row=failing_indices
+    )
+
+
+def test_validate():
+
+    p_death = pd.to_datetime(
```
tab1tha, line 71: I changed this name so that it fits python naming conventions. Typically, variable names should be made of lowercase separated by an underscore. This is called snake case.
Objects (e.g python classes) are named using camel case where the first letter of each word in multi-word variable name is capitalized.

PR #144, opened 2022-10-17
delete unnecessary files

tab1tha:
Deletes the code and files that no longer interact with the rest of the tool including.
- track_everything.py
- docker image of the .devcontainer file.
- requirements_dev.txt 
- CINXML_to_CSV.py which contains the former, inefficient method of performing the ingress process.


tab1tha:
Apparently, some of these files are necessary. The build process fails when they're absent.

PR #148, opened 2022-10-18
Added8905

PhilDeakin:
Added8905


tab1tha (approved):
This is good work.
- I removed the column names from strings as explained earlier.
- ran a formatter.
Thank you Phil !

PR #152, opened 2022-10-21
Rule_8840

willLPD2I:
closes #122


tab1tha:
Hi @willLPD2I, 
I changed the comments to fit your code so that your file can also be used as a template file since you are using it in the guide. 

Merging this in.

PR #153, opened 2022-10-21
use ReferenceDate in make_census_period

tab1tha:
A lot of the rules use the word `ReferenceDate` and not just `period of census`.
Although the `Year` can be used to generate the period of census, it makes more sense to use the column name that is more frequently associated with it.


PhilDeakin (approved):
This looks fine to me.

tab1tha:
Hi @PhilDeakin , 
The make_census_period function now wraps up your implementation.

The file changes here https://github.com/data-to-insight/CIN-validator/pull/153/files show how our two solutions have been merged into one.

PR #154, opened 2022-10-24
Rule8930 Closes #106

### LINKED ISSUE #106: Rule 8930 : Child Protection Plan End Date must fall within the census year
If \<CPPendDate\> (N00115) is present, then\<CPPendDate\> (N00115) must fall within [Period_of_Census] inclusive

amyrichards0n:
#106

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8930.py
+++ b/cin_validator/rules/cin2022_23/rule_8930.py
@@ -74,6 +74,9 @@ def test_validate():
             {
                 CPPendDate: "01/10/2022"
             },  # 2 fail: October 1st is after March 31st, 2022. It is out of range
+            {
+                CPPendDate: pd.NA
+            },  # 2 fail: October 1st is after March 31st, 2022. It is out of range
         ]
     )
 
--- a/cin_validator/rules/cin2022_23/rule_8930.py
+++ b/cin_validator/rules/cin2022_23/rule_8930.py
@@ -47,8 +47,7 @@ def validate(
 
     # If <CPPendDate> (N00115) is present, then<CPPendDate> (N00115) must fall within [Period_of_Census] inclusive
     failing_indices = df[
-        (df[CPPendDate] < collection_start)
-        | (df[CPPendDate] > reference_date)
+        (df[CPPendDate] < collection_start) | (df[CPPendDate] > reference_date)
     ].index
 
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8930.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # the make_census_period function generates the start and end date so that you don't have to do it each time.
+    collection_start, reference_date = make_census_period(ref_date_series)
+
+    # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
+
+    # If <CPPendDate> (N00115) is present, then<CPPendDate> (N00115) must fall within [Period_of_Census] inclusive
+    failing_indices = df[
+        (df[CPPendDate] < collection_start) | (df[CPPendDate] > reference_date)
+    ].index
```
tab1tha, line 51: Nice !

cin_validator/rules/cin2022_23/rule_8930.py
```diff
@@ -0,0 +1,107 @@
...
+        [
+            {
+                CPPendDate: "01/03/2019"
+            },  # 0 fail: March 1st is before April 1st, 2021. It is out of range
+            {
+                CPPendDate: "01/04/2021"
+            },  # 1 pass: April 1st is within April 1st, 2021 to March 31st, 2022.
+            {
+                CPPendDate: "01/10/2022"
+            },  # 2 fail: October 1st is after March 31st, 2022. It is out of range
```
tab1tha, line 76: You've checked for when the date is before, within and after the range.
Good test coverage. 

What if the date is not present at all?
You can a row where CPPendDate is pd.NA (that is, null)

tab1tha: When you add the row, run the rule and see what happens. Does the row with the missing value fail, pass or is it ignored?. @amyrichards0n

tab1tha (changes requested):

cin_validator/rules/cin2022_23/rule_8930.py
tab1tha: On second thought, let's do this in the future.

Since it is a little change compared to the work you've already done, I can make the change and merge it in.
See you tomorrow!.

PR #161, opened 2022-10-28
XMLtoCSV CLI command

willLPD2I:
Creates a CLI command to convert XMLs to CSV. Test with: python -m cin_validator xmltocsv /workspaces/CIN-validator/fake_data/CIN_Census_2021.xml

Needs to output as zip file eventually.

### DIFF
--- a/cin_validator/__main__.py
+++ b/cin_validator/__main__.py
@@ -79,5 +79,32 @@ def test_cmd(rule, ruleset):
     pytest.main(test_files)
 
 
+@cli.command(name="xmltocsv")
+@click.argument("filename")
+def cli_converter(filename: str):
+    """Converts XML to CSV at selected filepath"""
+    if Path(filename).exists():
+        fulltree = ET.parse(filename)
+        root = fulltree.getroot()
+        data_files = XMLtoCSV(root)
+        cin_tables_dict = {
+            "Header": data_files.Header,
+            "ChildIdentifiers": data_files.ChildIdentifiers,
+            "ChildCharacteristics": data_files.ChildCharacteristics,
+            "ChildProtectionPlans": data_files.ChildProtectionPlans,
+            "CINdetails": data_files.CINdetails,
+            "CINplanDates": data_files.CINplanDates,
+            "Reviews": data_files.Reviews,
+            "Section47": data_files.Section47,
+        }
+        for k, v in cin_tables_dict.items():
+            #  TODO output CSVs as a zip file
+            filepath = Path(f'output_csvs/{k}.csv')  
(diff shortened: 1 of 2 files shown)


cin_validator/ingress.py
```diff
@@ -142,11 +142,6 @@ def create_Header(self, header):
         collection_elements = ["Collection", "Year", "ReferenceDate"]
         header_dict = get_values(collection_elements, header_dict, collection_details)
```
tab1tha, line 144: Nice. 
You can also remove the lines where they are initialized in the class' __init__() above.

cin_validator/__main__.py
```diff
@@ -79,5 +79,32 @@ def test_cmd(rule, ruleset):
     pytest.main(test_files)
 
 
+@cli.command(name="xmltocsv")
+@click.argument("filename")
+def cli_converter(filename: str):
```
tab1tha, line 84: @click.argument("filename", type=click.Path(), required=True)

This works also.

cin_validator/__main__.py
```diff
@@ -79,5 +79,32 @@ def test_cmd(rule, ruleset):
...
+            "ChildProtectionPlans": data_files.ChildProtectionPlans,
+            "CINdetails": data_files.CINdetails,
+            "CINplanDates": data_files.CINplanDates,
+            "Reviews": data_files.Reviews,
+            "Section47": data_files.Section47,
+        }
+        for k, v in cin_tables_dict.items():
+            #  TODO output CSVs as a zip file
+            filepath = Path(f'output_csvs/{k}.csv')  
+            filepath.parent.mkdir(parents=True, exist_ok=True)
```
tab1tha: Good work, Will !

cin_validator/__main__.py
```diff
@@ -79,5 +79,32 @@ def test_cmd(rule, ruleset):
...
+        cin_tables_dict = {
+            "Header": data_files.Header,
+            "ChildIdentifiers": data_files.ChildIdentifiers,
+            "ChildCharacteristics": data_files.ChildCharacteristics,
+            "ChildProtectionPlans": data_files.ChildProtectionPlans,
+            "CINdetails": data_files.CINdetails,
+            "CINplanDates": data_files.CINplanDates,
+            "Reviews": data_files.Reviews,
+            "Section47": data_files.Section47,
+        }
```
tab1tha, line 99: We will think of a way to do this without having to spell out all the column names one after the other. It could mean extending the DataContainer mapping used in the run-all CLI command such that it has a keys() method that returns the keys of the mapping. 

Let's put a pin on that.

tab1tha (approved):
It works.
This so cool.

PR #167, opened 2022-10-31
Rule 2885: Example of type2.

tab1tha:
Implements rule 2885 as a demonstration of what a type2 rule looks like if 3 tables are involved. 
Type 2 rules typically involved 2 tables so they'll typically require less lines of code than this rule.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_2885.py
+++ b/cin_validator/rules/cin2022_23/rule_2885.py
@@ -16,6 +16,7 @@ CPPstartDate = ChildProtectionPlans.CPPstartDate
 # is it necessary to get LAchildID per table?
 LAchildID = ChildProtectionPlans.LAchildID
 CINdetailsID = ChildProtectionPlans.CINdetailsID
+# TODO is it redundant to have both of them here?
 DateOfInitialCPC = Section47.DateOfInitialCPC
 DateOfInitialCPC = CINdetails.DateOfInitialCPC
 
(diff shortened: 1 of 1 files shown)


willLPD2I (approved):

cin_validator/rules/cin2022_23/rule_2885.py
```diff
@@ -0,0 +1,360 @@
...
+    ],  # TODO How can we indicate that the DateOfInitialCPC comes from both tables. Is it necessary?
+)
+def validate(
+    data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
+):
+    # PREPARING DATA
+
+    # Replace ChildProtectionPlans with the name of the table you need.
+    df_cpp = data_container[ChildProtectionPlans].copy()
+    # TODO Is it necessary to copy? How else can we ensure that the original data is not changed in each rule.?.
```
willLPD2I, line 49: It's inside a function so it should change the original data unless you return it, right?

tab1tha: That's a fair argument.

willLPD2I: I think it's safest to use .loc[] in instances like this, right?

tab1tha: From what I know, .loc tends to be used with DataFrame-like structures. Is it used with dictionary-like structures too?.
data_container is a dictionary-like mapping so it is flat, no rows and columns.

tab1tha: I've converted this to draft so that I don't mistakenly merge it. There are a few changes that have to be made.

willLPD2I: Oh I'm a fool, you're right, I just assumed it was a df when clearly it isn't

PR #168, opened 2022-10-31
Rule8810

### LINKED ISSUE #48: Rule 8810 : A CIN case cannot have a Reason for Closure without a CIN Closure Date
If \<ReasonForClosure\> (N00103) is present then \<CINclosureDate\> (N00102) must also be present

willLPD2I:
closes #48

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8810.py
+++ b/cin_validator/rules/cin2022_23/rule_8810.py
@@ -54,7 +54,9 @@ def validate(
 
     # Replace CPPstartDate and CPPendDate below with the columns concerned in your rule.
     link_id = tuple(
-        zip(df_issues[LAchildID], df_issues[ReasonForClosure], df_issues[CINclosureDate])
+        zip(
+            df_issues[LAchildID], df_issues[ReasonForClosure], df_issues[CINclosureDate]
+        )
     )
     df_issues["ERROR_ID"] = link_id
     df_issues = df_issues.groupby("ERROR_ID")["ROW_ID"].apply(list).reset_index()
@@ -92,7 +94,7 @@ def test_validate():
                 "LAchildID": "child4",
                 "ReasonForClosure": pd.NA,
                 "CINclosureDate": "25/05/2000",
-            },  
+            },
             {
                 "LAchildID": "child5",
                 "ReasonForClosure": pd.NA,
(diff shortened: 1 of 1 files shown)


willLPD2I:
I've updated the link_id to a non null, hope it's right now!

cin_validator/rules/cin2022_23/rule_8810.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def test_validate():
+    # Create some sample data such that some values pass the validation and some fail.
+    fake_data_frame = pd.DataFrame(
+        [
+            {
+                "LAchildID": "child1",
+                "CINdetailsID": "CINID1",
+                "ReasonForClosure": "26/05/2000",
```
tab1tha, line 75: This does work. However, the data values for ReasonForClosure are more relatable if they are strings. The reason cannot be a date. Date columns tend to have the word "date" in their name.

tab1tha: You can use random strings like "aaa", "bbb", "abc" ....

cin_validator/rules/cin2022_23/rule_8810.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # SUBMIT ERRORS
+    # Generate a unique ID for each instance of an error. In this case,
+    # - If only LAchildID is used as an identifier, multiple instances of the error on a child will be understood as 1 instance.
+    # We don't want that because in reality, a child can have multiple instances of an error.
+    # - If we use the LAchildID-CPPstartDate combination, that artificially cancels out the instances where a start date repeats for the same child.
+    # Another rule checks for that condition. Not this one.
+    # - It is very unlikely that a combination of LAchildID-CPPstartDate-CPPendDate will repeat in the DataFrame.
+    # Hence, it can be used as a unique identifier of the row.
+
```
tab1tha, line 55: You can remove this comment block. This explanation is only relevant for the first rule that it explained.

tab1tha (changes requested):
After the changes suggested, everything else is good to go.

Nice work !

cin_validator/rules/cin2022_23/rule_8810.py
willLPD2I: HAha, I was just trying to save time! I'll do this now

willLPD2I: done

tab1tha (approved):

PR #169, opened 2022-10-31
Added rule 8750

### LINKED ISSUE #22: Rule 8750 : Gender must equal 0 for an unborn child
If \<ExpectedPersonBirthDate\> (N00098) is present and \<PersonBirthDate\> (N00066) is blank then \<GenderCurrent\> (N00065) must equal “0”

SLornieCYC:
Closes #22

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8750.py
+++ b/cin_validator/rules/cin2022_23/rule_8750.py
@@ -73,6 +73,7 @@ def validate(
         row_df=df_issues,
     )
 
+
 def test_validate():
     # Create some sample data such that some values pass the validation and some fail.
     child_identifiers = pd.DataFrame(
--- a/cin_validator/rules/cin2022_23/rule_8750.py
+++ b/cin_validator/rules/cin2022_23/rule_8750.py
@@ -65,7 +65,6 @@ def validate(
         )
     )
     df_issues["ERROR_ID"] = link_id
-    print ('df_issues with link_id \n', df_issues)
     df_issues = df_issues.groupby("ERROR_ID")["ROW_ID"].apply(list).reset_index()
     # Ensure that you do not change the ROW_ID, and ERROR_ID column names which are shown above. They are keywords in this project.
     rule_context.push_type_1(
@@ -73,7 +72,6 @@ def validate(
         columns=[GenderCurrent, PersonBirthDate, ExpectedPersonBirthDate],
         row_df=df_issues,
     )
-    print ('df_issues after groupby \n', df_issues)
 
 def test_validate():
     # Create some sample data such that some values pass the validation and some fail.
@@ -157,23 +155,21 @@ def test_validate():
             {
                 "ERROR_ID": (
                     "child4",
-                    "2",
-                    pd.to_datetime("25/05/2000", format="%d/%m/%Y", errors="coerce"),
-                ),
-                "ROW_ID": [4],
-            },
-            {
-                "ERROR_ID": (
-                    "child4",
                     "1",
                     pd.to_datetime("25/05/2000", format="%d/%m/%Y", errors="coerce"),
                 ),
                 "ROW_ID": [5],
             },
+            {
+                "ERROR_ID": (
+                    "child4",
+                    "2",
+                    pd.to_datetime("25/05/2000", format="%d/%m/%Y", errors="coerce"),
+                ),
+                "ROW_ID": [4],
+            },
         ]
     )
-    print ('issue_rows \n', issue_rows)
-    print ('expected_df \n', expected_df)
     assert issue_rows.equals(expected_df)
 
     # Check that the rule definition is what you wrote in the context above.
--- a/cin_validator/rules/cin2022_23/rule_8750.py
+++ b/cin_validator/rules/cin2022_23/rule_8750.py
@@ -1,0 +1,166 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import CINTable, RuleContext, rule_definition
+from cin_validator.test_engine import run_rule
+
+# Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
+
+ChildIdentifiers = CINTable.ChildIdentifiers
+LAchildID = ChildIdentifiers.LAchildID
+PersonBirthDate = ChildIdentifiers.PersonBirthDate
+ExpectedPersonBirthDate = ChildIdentifiers.ExpectedPersonBirthDate
+GenderCurrent = ChildIdentifiers.GenderCurrent
+
+# define characteristics of rule
+@rule_definition(
+    # write the rule code here, in place of 8500
+    code=8750,
+    # replace ChildIdentifiers with the value in the module column of the excel sheet corresponding to this rule .
+    module=CINTable.ChildIdentifiers,
+    # replace the message with the corresponding value for this rule, gotten from the excel sheet.
+    message="Gender must equal 0 for an unborn child",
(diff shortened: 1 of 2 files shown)


cin_validator/rules/cin2022_23/rule_8750.py
```diff
@@ -0,0 +1,166 @@
...
+# define characteristics of rule
+@rule_definition(
+    # write the rule code here, in place of 8500
+    code=8750,
+    # replace ChildIdentifiers with the value in the module column of the excel sheet corresponding to this rule .
+    module=CINTable.ChildIdentifiers,
+    # replace the message with the corresponding value for this rule, gotten from the excel sheet.
+    message="Gender must equal 0 for an unborn child",
+    # The column names tend to be the words within the < > signs in the github issue description.
+    affected_fields=[GenderCurrent, ExpectedPersonBirthDate],
```
willLPD2I, line 25: I think this should also include PersonBirthDate?

SLornieCYC: It feels excessive to include and highlight all three fields for this error.  On reflection I think I should probably change it to GenderCurrent and **PersonBirthDate** instead of GenderCurrent and ExpectedPersonBirthDate, on the basis that the primary identifier of "unborn" is lack of DOB.

image

cin_validator/rules/cin2022_23/rule_8750.py
```diff
@@ -0,0 +1,166 @@
...
+    # Generate a unique ID for each instance of an error. In this case,
+    # - If only LAchildID is used as an identifier, multiple instances of the error on a child will be understood as 1 instance.
+    # We don't want that because in reality, a child can have multiple instances of an error.
+    # - If we use the LAchildID-CPPstartDate combination, that artificially cancels out the instances where a start date repeats for the same child.
+    # Another rule checks for that condition. Not this one.
+    # - It is very unlikely that a combination of LAchildID-CPPstartDate-CPPendDate will repeat in the DataFrame.
+    # Hence, it can be used as a unique identifier of the row.
+
+    # Replace CPPstartDate and CPPendDate below with the columns concerned in your rule.
+    link_id = tuple(
```
SLornieCYC, line 60: @willLPD2I Not quite sure I understand this bit of the guidance/coding.  In this instance do I need to include additional fields beyond LAchildID for link_id? Theoretically ChildIndentifiers is one-row-per-child.

cin_validator/rules/cin2022_23/rule_8750.py
willLPD2I: So, it's to do with which cells are highlighted on the front end and I just want to check with Tambe which method will give the expected behaviour! Obviously, there shouldn't be any instances where someone has an expected birth date and a DoB, but it's more to do with what the front end expects for the outputs.

tab1tha:
@SLornieCYC checking in.
When you have done a rule and the logic works, are you keen on doing the finishing touches or do you prefer that we do those and merge?. What works for you?

I'm happy either way.

SLornieCYC:
hi @tab1tha I think I'd like to finish it off myself.  It'll be good practice and makes sure I properly take note of what needs to be done!

tab1tha:
Hi @SLornieCYC 
This pull request has been inactive for a while. It is possible that since many comments were made, what needs to be changed in order for the pull request to be approved might not be totally clear. So I have resolved the large conversation and now I will go through and point out the exact parts that need changing. I hope that helps. I'm also happy to hop on a call if it doesn't.

cin_validator/rules/cin2022_23/rule_8750.py
```diff
@@ -0,0 +1,166 @@
...
+        zip(
+            df_issues[LAchildID],
+            df_issues[GenderCurrent],
+            df_issues[PersonBirthDate],
+        )
+    )
```
tab1tha, line 66: 2/3 
As mentioned in the note, the ERROR_ID doesn't need to contain PersonBirthDate. It can be formed from a combination of 
- LAchildID and ExpectedPersonBirthDate (preferred)
- LAchildID and GenderCurrent
but not the three. The former combination is preferred because it fits the guideline of creating ERROR_ID from the "condition" columns. However, the latter combination is also appropriate in this case.

cin_validator/rules/cin2022_23/rule_8750.py
```diff
@@ -0,0 +1,166 @@
...
+    # lOGIC
+    # Implement rule logic as described by the Github issue.
+    # Put the description as a comment above the implementation as shown.
+
+    # If <ExpectedPersonBirthDate> (N00098) is present and <PersonBirthDate> (N00066) is blank then <GenderCurrent> (N00065) must equal “0”
+    condition = (
+        df[PersonBirthDate].isna()
+        & df[ExpectedPersonBirthDate].notna()
+        & (df[GenderCurrent].astype(str) != "0")
+    )
```
tab1tha, line 46: This is good !

cin_validator/rules/cin2022_23/rule_8750.py
```diff
@@ -0,0 +1,166 @@
...
+                "ERROR_ID": (
+                    "child4",
+                    "2",
+                    pd.NaT,
+                ),
+                "ROW_ID": [4],
+            },
+        ]
+    )
+    assert issue_rows.equals(expected_df)
```
tab1tha, line 160: 3/3
expected_df needs to be changed to reflect the updated ERROR_ID combination.

tab1tha (changes requested):

cin_validator/rules/cin2022_23/rule_8750.py
SLornieCYC: Afraid I still don't quite understand this.  If we assume that _somehow_ an LA has loaded multiple rows for the same child into the one-row-per-child ChildIdentifiers table, you could end up with, for example:

| ChildID | ExpectedDOB | ActualDOB | Gender |
| -------- | -------------- | ------------- | -------- |
| Child 1 | 10/10/2020 | NULL | 0 |
| Child 1 | 10/10/2020 | NULL | 1 |
| Child 1 | 10/10/2020 | NULL | 2 |
| Child 1 | 20/10/2020 | NULL | 2 |

(Probably not an exhaustive list)

That's potentially multiple in-error rows with the same EDD and different gender and multiple rows with different EDD and same gender all which would need to be listed out separately.

From the note that makes me think this would need to zip all three fields (not including ActualDOB) to cover all possibilities for distinct rows?

SLornieCYC:
@tab1tha For some reason the row order in `df_issues` changes between adding `link_id` and pushing the errors here, which means they turn up in reverse order compared to what I expect to record in `expected_df`.  Should I just accept this and re-order my `expected_df` to match, or is there another issue here?

I've included print statements in my latest commit for testing purposes, which output like this:

image

SLornieCYC:
Hi @tab1tha, I've reversed the order of my expected test failures to match the output from passing the test data through the rule.  As discussed https://github.com/data-to-insight/CIN-validator/pull/169#discussion_r1019029881 I think all three fields (excluding ActualDOB) are required in link_id for distinctness.

tab1tha (approved):
Good work @SLornieCYC 

We're happy to merge in your first pull request to the CIN-validator repo, hence I get to say:
Welcome to be team @SLornieCYC !

tab1tha:
> Hi @tab1tha, I've reversed the order of my expected test failures to match the output from passing the test data through the rule. As discussed [#169 (comment)](https://github.com/data-to-insight/CIN-validator/pull/169#discussion_r1019029881) I think all three fields (excluding ActualDOB) are required in link_id for distinctness.

True. That works.

PR #170, opened 2022-11-01
Added8630

PhilDeakin:
Added8630

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8630.py
+++ b/cin_validator/rules/cin2022_23/rule_8630.py
@@ -1,0 +1,146 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import rule_definition, CINTable, RuleContext
+from cin_validator.rule_engine import IssueLocator
+from cin_validator.test_engine import run_rule
+
+CINdetails = CINTable.CINdetails
+CINreferralDate = CINdetails.CINreferralDate
+CINclosureDate = CINdetails.CINclosureDate
+LAchildID = CINdetails.LAchildID
+
+
+# define characteristics of rule
+@rule_definition(
+    code="8630",
+    module=CINTable.CINdetails,
+    message="CIN Closure Date is before CIN Referral Date for the same CIN episode",
+    affected_fields=[CINreferralDate, CINclosureDate],
+)
+def validate(
+    data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
+):
+    df = data_container[CINdetails]
+
+    """
+    If present <CINclosureDate> (N00102) must be on or after <CINreferralDate> (N00100) in the same <CINdetails> instance
+    """
+
+    df.index.name = "ROW_ID"
+
+    df.query(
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8630.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    If present <CINclosureDate> (N00102) must be on or after <CINreferralDate> (N00100) in the same <CINdetails> instance
+    """
+
+    df.index.name = "ROW_ID"
+
+    df.query(
+        "(CINclosureDate < CINreferralDate) and CINclosureDate.notna() and CINreferralDate.notna()",
+        inplace=True,
+    )
```
tab1tha, line 36: Ouuu this is interesting. I like it.
df.query() is relevant for these usecases.

Sadly, it requires that the column names are put in strings. This time, we've put in some measures to prevent the column names from being misspelled or mapped to the wrong table. The changes also enable the column names to be changed throughout the file from a central location (that is, without needing to find and change them in each file manually). If the column names are put in strings, these features are masked so we need to keep the standard column names out of strings.

The only column names that are used in string form should be the additional columns that are created as part of implementing the rule e.g ROW_ID, ERROR_ID, and any other columns you create as part of your logic.

Still, it is really nice to see you implement something you've learned. @PhilDeakin

tab1tha (changes requested):
Since the changes are minor, I'll push them and merge as you said you preferred.

cin_validator/rules/cin2022_23/rule_8630.py
tab1tha: I noticed you used "and" (boolean comparator) in place of "&" (bitwise comparator).
Though that works in this test, it can produce unexpected results. 
If you have to compare the truth value of a structure to that of another structure, you can use "and". However, if you have to compare the elements of those structures to each other (in this case, columns in each row of the dataframe), "&" is more appropriate. 

You can read more about the differences here: https://stackoverflow.com/a/22647006/11044184

tab1tha (approved):

tab1tha:
If you are curious about where the column names are enforced, it is in the CINTable class in the api.py file of the rule engine.

Also, if you are curious about how it is possible to change the name of a variable from a central location, (Don't do this on the project itself, or do it and reverse the changes), this is how it is done:
- If you right-click on a keyword e.g a column name, a large pop up appears.
- Clicking on the option "Rename Symbol" allows you to change the name of the element you clicked on, throughout the repo.
This is useful if the government guidelines ever change and a change of column names is required.

PhilDeakin:
Going on "Change All Occurrences" also grabs occurrences within strings, but I'm happy to go along with what you've suggested.

cin_validator/rules/cin2022_23/rule_8630.py
PhilDeakin: @tab1tha I presume functional strings would be fine e.g.
df.query(f"({CINclosureDate} < {CINreferralDate}) & {CINclosureDate}.notna() & {CINreferralDate}.notna()")

PR #171, opened 2022-11-01
Rule8841

### LINKED ISSUE #108: Rule 8841 : The review date cannot be on the same day or before the Child protection Plan start date.
Within a \<ChildProtectionPlans\> group, there should be no \<CPPreviewDate\> (N00116) that is the same as or before the \<CPPstartDate\> (N00105)

willLPD2I:
closes #108 
May have to be saved as a draft as it's based on Tambes rule 2885 which is also a draft.
It's an easier rule as it only contains 2 tables and no census periods so may make a better template than 2885 in some cases, and a worse template in others.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8841.py
+++ b/cin_validator/rules/cin2022_23/rule_8841.py
@@ -9,16 +9,17 @@ from cin_validator.utils import make_census_period
 # Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
 
 ChildProtectionPlans = CINTable.ChildProtectionPlans
-CPPstartDate =  ChildProtectionPlans.CPPstartDate
+CPPstartDate = ChildProtectionPlans.CPPstartDate
 LAchildID = ChildProtectionPlans.LAchildID
 CPPID_CPP = ChildProtectionPlans.CPPID
+
 Reviews = CINTable.Reviews
-CPPreviewDate = Reviews.CPPreviewDate 
+CPPreviewDate = Reviews.CPPreviewDate
 CPPID_reviews = Reviews.CPPID
 
 # define characteristics of rule
 @rule_definition(
-    # write the rule code here, in place of 2885
+    # write the rule code here
     code=8841,
     # replace ChildProtectionPlans with the value in the module column of the excel sheet corresponding to this rule .
     # Note that even if multiple tables are involved, one table will be named in the module column.
@@ -29,7 +30,7 @@ CPPID_reviews = Reviews.CPPID
     affected_fields=[
         CPPstartDate,
         CPPreviewDate,
-    ],  # TODO How can we indicate that the DateOfInitialCPC comes from both tables. Is it necessary?
+    ],
 )
 def validate(
     data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
@@ -40,15 +41,12 @@ def validate(
     df_cpp = data_container[ChildProtectionPlans].copy()
     df_reviews = data_container[Reviews].copy()
 
-
-
     # Before you begin, rename the index so that the initial row positions can be kept intact.
     df_cpp.index.name = "ROW_ID"
     df_reviews.index.name = "ROW_ID"
 
     # Resetting the index causes the ROW_IDs to become columns of their respective DataFrames
     # so that they can come along when the merge is done.
-    # TODO summarise with a for loop? e.g for df in [df_cpp, df_47, df_cin]
     df_cpp.reset_index(inplace=True)
     df_reviews.reset_index(inplace=True)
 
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8841.py
```diff
@@ -0,0 +1,281 @@
...
+            ChildProtectionPlans: sample_cpp,
+            Reviews: sample_reviews,
+            ChildProtectionPlans: sample_cpp,
+        },
+    )
+
+    # Use .type2_issues to check for the result of .push_type2_issues() which you used above.
+    issues_list = result.type2_issues
+    assert (
+        len(issues_list) == 2
```
willLPD2I, line 211: Tambe, I wasn't sure about this line

willLPD2I:
@tab1tha can you review this type 2 when you get a chance, please?

cin_validator/rules/cin2022_23/rule_8841.py
```diff
@@ -0,0 +1,279 @@
...
+    #  Create dataframes which only have rows with CP plans, and which should have one plan per row.
+    df_cpp = df_cpp[df_cpp[CPPstartDate].notna()]
+    df_reviews = df_reviews[df_reviews[CPPreviewDate].notna()]
+
+    #  Merge tables to get corresponding CP plan group and reviews
+    df_merged = df_cpp.merge(
+        df_reviews,
+        left_on=["CPPID", "LAchildID"],
+        right_on=["CPPID", "LAchildID"],
+        how="left",
```
tab1tha, line 69: nice. 
"within a child protection plans group..." means that we should only concern ourselves with the values that are present in the child protection plans table and your decision to merge left does just that !

cin_validator/rules/cin2022_23/rule_8841.py
```diff
@@ -0,0 +1,279 @@
...
+
+    #  Create dataframes which only have rows with CP plans, and which should have one plan per row.
+    df_cpp = df_cpp[df_cpp[CPPstartDate].notna()]
+    df_reviews = df_reviews[df_reviews[CPPreviewDate].notna()]
+
+    #  Merge tables to get corresponding CP plan group and reviews
+    df_merged = df_cpp.merge(
+        df_reviews,
+        left_on=["CPPID", "LAchildID"],
+        right_on=["CPPID", "LAchildID"],
```
tab1tha, line 68: Almost there.

If you check the sample dataframes file (https://github.com/data-to-insight/CIN-validator/blob/main/cin_validator/demo_ingress_class.ipynb) or see the table column lists in api.py, you'll notice that in addition to "CPPID" and  "LAchildID", these tables also have to be grouped by CINdetailsID. 

So that has to be included in your join too if not two CPP groups that belong to different CINdetails groups but have the same ID would be seen as if they're the same.

cin_validator/rules/cin2022_23/rule_8841.py
```diff
@@ -0,0 +1,279 @@
...
+        how="left",
+        suffixes=("_cpp", "_reviews"),
+    )
+
+    #  Get rows where CPPreviewDate is less than or equal to CPPstartDate
+    condition = df_merged[CPPreviewDate] <= df_merged[CPPstartDate]
+    df_merged = df_merged[condition].reset_index()
```
tab1tha, line 75: Great. 

The rest is good to go !
Thanks Will.

cin_validator/rules/cin2022_23/rule_8841.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def test_validate():
+    # Create some sample data such that some values pass the validation and some fail.
+    sample_cpp = pd.DataFrame(
+        [
+            {
+                "LAchildID": "child1",
+                "CPPstartDate": "26/05/2000",  # Fails as dates are the same
+                "CPPID": "cinID1",
```
tab1tha, line 115: CINdetailsID will have to be considered in the test data also.

cin_validator/rules/cin2022_23/rule_8841.py
```diff
@@ -0,0 +1,281 @@
...
+            Reviews: sample_reviews,
+            ChildProtectionPlans: sample_cpp,
+        },
+    )
+
+    # Use .type2_issues to check for the result of .push_type2_issues() which you used above.
+    issues_list = result.type2_issues
+    assert (
+        len(issues_list) == 2
+    )  #  Tambe, am I right in thinking this is the number of tables with issues?
```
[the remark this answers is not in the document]
tab1tha: Yes you are. 
Issues_list contains all the tables flagged. So it's length is the number of tables involved in the rule.

tab1tha (changes requested):

tab1tha (approved):

PR #172, opened 2022-11-02
Rule 8896: example of type3

tab1tha:
Type3 rules tend to involve 1 column. They define things that should not happen in a group.
Examples of groups are CINdetails groups, CPP groups and so on.


tab1tha:
@willLPD2I could you review this too?

willLPD2I (approved):

PR #183, opened 2022-11-09
Fixes print statements, adds date input flexibility for reference date

willLPD2I:
Some rules had print statement sin which printed to the console when running through the CLI

Reference dates could be formatted in one of two ways, this allows both within the make_census_period function. It's done inside that function for now as it doesn't seem like it's needed anywhere else, and it means we wouldn't have to change how any rules are coded. Potential future work to DRY it up?

### DIFF
--- a/cin_validator/__main__.py
+++ b/cin_validator/__main__.py
@@ -48,7 +48,7 @@ def run_all(filename: str, ruleset):
 
         ctx = RuleContext(rule)
         rule.func(data_files, ctx)
-        #  TODO determine what to use for the group_keys argument 
+        #  TODO determine what to use for the group_keys argument
         print(rule.code, len(list(ctx.issues)))
 
 
--- a/cin_validator/__main__.py
+++ b/cin_validator/__main__.py
@@ -47,9 +47,14 @@ def run_all(filename: str, ruleset):
     for rule in registry:
 
         ctx = RuleContext(rule)
+
         rule.func(data_files, ctx)
         #  TODO determine what to use for the group_keys argument
-        print(rule.code, len(list(ctx.issues)))
+        if len(list(ctx.issues)) == 0:
+            print(rule.code, len(list(ctx.issues)))
+        else:
+            for i in range(len(list(ctx.issues))):
+                print(rule.code, list(ctx.issues)[i], rule.message)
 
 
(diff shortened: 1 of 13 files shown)


willLPD2I:
OH, I forgot to put this in the PR, this also prints errors line by line to the console. It's an early feature which probably need expansion and refinement.

cin_validator/__main__.py
```diff
@@ -47,8 +47,14 @@ def run_all(filename: str, ruleset):
...
         ctx = RuleContext(rule)
+
         rule.func(data_files, ctx)
-        print(rule.code, len(list(ctx.issues)))
+        #  TODO determine what to use for the group_keys argument
+        if len(list(ctx.issues)) == 0:
+            print(rule.code, len(list(ctx.issues)))
+        else:
+            for i in range(len(list(ctx.issues))):
+                print(rule.code, list(ctx.issues)[i], rule.message)
```
tab1tha, line 57: Accumulate this into an object. 
Then print the object.

cin_validator/utils.py
```diff
@@ -21,13 +30,14 @@ def make_census_period(reference_date):
...
     # the ReferenceDate value is always the collection_end date
-    collection_end = pd.to_datetime(reference_date, format="%d/%m/%Y")
+    collection_end = make_date(reference_date)
+
     # the collection start is the 1st of April of the previous year.
     collection_start = (
-        pd.to_datetime(reference_date, format="%d/%m/%Y")
-        - pd.DateOffset(years=1)
-        + pd.DateOffset(days=1)
+        make_date(reference_date) - pd.DateOffset(years=1) + pd.DateOffset(days=1)
```
tab1tha, line 40: This no longer needs to be surrounded by brackets. the make_date statement can exist without the brackets around it on lines 39 and 41.

tab1tha (changes requested):

cin_validator/utils.py
willLPD2I: black actually automatically adds the brackets, it's to keep line length within 79 characters believe

willLPD2I:
Group keys all set = False, only needs to be done in the creation of df_issues to prevent warnings

PR #184, opened 2022-11-09
Added4015

PhilDeakin:
Added4015

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4015.py
+++ b/cin_validator/rules/cin2022_23/rule_4015.py
@@ -1,0 +1,205 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import CINTable, RuleContext, rule_definition
+from cin_validator.test_engine import run_rule
+from cin_validator.utils import make_census_period
+
+# Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
+
+CINplanDates = CINTable.CINplanDates
+CINPlanStartDate = CINplanDates.CINPlanStartDate
+LAchildID = CINplanDates.LAchildID
+
+CINdetails = CINTable.CINdetails
+CINreferralDate = CINdetails.CINreferralDate
+
+# define characteristics of rule
+@rule_definition(
+    code=4015,
+    # Note that even if multiple tables are involved, one table will be named in the module column.
+    module=CINTable.CINplanDates,
+    message="The CIN Plan start date cannot be before the referral date",
+    affected_fields=[
+        CINPlanStartDate,
+        CINreferralDate,
+    ],
+)
+def validate(
+    data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
+):
+    """
+    Where present, the <CINPlanStartDate> (N00689) must be on or after the <CINReferralDate> (N00100)
+    """
+    df_cindetail = data_container[CINdetails].copy()
+    df_cinplan = data_container[CINplanDates].copy()
+
+    print(df_cindetail)
+    print(df_cinplan)
+
+    df_cindetail.index.name = "ROW_ID"
+    df_cinplan.index.name = "ROW_ID"
+
+    df_cindetail.reset_index(inplace=True)
+    df_cinplan.reset_index(inplace=True)
+
+    df_cindetail = df_cindetail[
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_4015.py
```diff
@@ -0,0 +1,205 @@
...
+    df_cindet_issues = (
+        df_cinplan.merge(
+            df_merged, how="inner", left_on="ROW_ID", right_on="ROW_ID_plan"
+        )
+        .groupby("ERROR_ID")["ROW_ID"]
+        .apply(list)
+        .reset_index()
+    )
+
+    df_cindet_issues.sort_values(["ROW_ID"], ignore_index=True, inplace=True)
```
PhilDeakin, line 80: This line matters... the " assert issue_rows.equals(expected_df) " near the bottom can fail otherwise.

cin_validator/rules/cin2022_23/rule_4015.py
```diff
@@ -0,0 +1,205 @@
...
+    df_cinplan_issues = (
+        df_cindetail.merge(
+            df_merged, how="inner", left_on="ROW_ID", right_on="ROW_ID_det"
+        )
+        .groupby("ERROR_ID")["ROW_ID"]
+        .apply(list)
+        .reset_index()
+    )
+
+    df_cinplan_issues.sort_values(["ROW_ID"], ignore_index=True, inplace=True)
```
PhilDeakin, line 91: This line matters... the " assert issue_rows.equals(expected_df) " near the bottom can fail otherwise.

cin_validator/rules/cin2022_23/rule_4015.py
```diff
@@ -0,0 +1,205 @@
...
+def validate(
+    data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
+):
+    """
+    Where present, the <CINPlanStartDate> (N00689) must be on or after the <CINReferralDate> (N00100)
+    """
+    df_cindetail = data_container[CINdetails].copy()
+    df_cinplan = data_container[CINplanDates].copy()
+
+    print(df_cindetail)
```
willLPD2I, line 38: Try not to leave print statements in!

cin_validator/rules/cin2022_23/rule_4015.py
```diff
@@ -0,0 +1,205 @@
...
+    df_cindetail.index.name = "ROW_ID"
+    df_cinplan.index.name = "ROW_ID"
+
+    df_cindetail.reset_index(inplace=True)
+    df_cinplan.reset_index(inplace=True)
+
+    df_cindetail = df_cindetail[
+        ["ROW_ID", "LAchildID", "CINdetailsID", "CINreferralDate"]
+    ]
+    df_cinplan = df_cinplan[["ROW_ID", "LAchildID", "CINdetailsID", "CINPlanStartDate"]]
```
willLPD2I, line 50: I think lines 47-50 here are redundant?

willLPD2I (changes requested):
Some small changes needed. I am curious about the sort_values lines, I can see they're needed, I just wonder why when no other type_2 rule needs them to my knowledge, and I wonder if they'll have an effect down the road, or if we've got a bug somewhere that comes in before this rule is run that causes the issue.

willLPD2I:
@tab1tha just to check, we agreed that the sorting in all likelihood wasn't going to be a problem but we might see a) why it's needed and b) if we can do it in validate instead in case it messes with something unanticipated? So I might put this on hold?

tab1tha:
I have come back to this and haven't found a way to isolate the sort from the validate function while maintaining a logical prediction of what I expect the result to look like. 
This is because sorting is not needed on the data passed into the function (which the `test_validate()` functions has access to) but on the data generated within the validate function (which only the `validate() `function has access to)

tab1tha (approved):

willLPD2I (approved):

tab1tha:
Oops, I take that back.
I have just found a way to put it in the test_validate function. 
I'll push it since it is a small change @PhilDeakin

tab1tha:
I added this line `issue_rows.sort_values(["ROW_ID"], ignore_index=True, inplace=True)` just before this line `assert issue_rows.equals(expected_df)` and it worked. @PhilDeakin 

Merging this in...

PR #189, opened 2022-11-11
Added rule 8740

### LINKED ISSUE #91: Rule 8740 : For a Section 47 Enquiry that has not held the Initial Child Protection Conference by the end of the census year, the start date must fall within the census year
If a \<Section47\> group does not contain the \<DateOfInitialCPC\> (N00110) and \<ICPCnotRequired\> (N00111) is false then the \<S47ActualStartDate\> (N00148) must be on or between [Start_Of_Census_Year] and \<ReferenceDate\> (N00603)

SLornieCYC:
Closes #91

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8740.py
+++ b/cin_validator/rules/cin2022_23/rule_8740.py
@@ -1,0 +1,201 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import CINTable, RuleContext, rule_definition
+from cin_validator.test_engine import run_rule
+from cin_validator.utils import make_census_period
+
+# Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
+
+Section47 = CINTable.Section47
+LAchildID = Section47.LAchildID
+S47ActualStartDate = Section47.S47ActualStartDate
+DateOfInitialCPC = Section47.DateOfInitialCPC
+ICPCnotRequired = Section47.ICPCnotRequired
+
+Header = CINTable.Header
+ReferenceDate = Header.ReferenceDate
+
+# define characteristics of rule
+@rule_definition(
+    # write the rule code here, in place of 8500
+    code=8740,
+    # replace ChildIdentifiers with the value in the module column of the excel sheet corresponding to this rule .
+    module=CINTable.Section47,
+    # replace the message with the corresponding value for this rule, gotten from the excel sheet.
+    message="For a Section 47 Enquiry that has not held the Initial Child Protection Conference by the end of the census year, the start date must fall within the census year",
+    # The column names tend to be the words within the < > signs in the github issue description.
+    affected_fields=[S47ActualStartDate, DateOfInitialCPC, ICPCnotRequired],
+)
+def validate(
+    data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
+):
+    # PREPARING DATA
+
+    # Replace ChildIdentifiers with the name of the table you need.
+    df = data_container[Section47]
+    # Before you begin, rename the index so that the initial row positions can be kept intact.
+    df.index.name = "ROW_ID"
+
+    # ReferenceDate exists in the heder table so we get header table too.
+    df_ref = data_container[Header]
+    ref_date_series = df_ref[ReferenceDate]
+
+    # the make_census_period function generates the start and end date so that you don't have to do it each time.
+    collection_start, reference_date = make_census_period(ref_date_series)
+
+    # lOGIC
+    # Implement rule logic as described by the Github issue.
+    # Put the description as a comment above the implementation as shown.
+    print ('\n initial df \n', df)
+    # If a <Section47> group does not contain the <DateOfInitialCPC> (N00110) and <ICPCnotRequired> (N00111) is false
+    # then the <S47ActualStartDate> (N00148) must be on or between [Start_Of_Census_Year] and <ReferenceDate> (N00603)
+    condition = (
+        df[DateOfInitialCPC].isna() 
+        & (df[ICPCnotRequired].astype(str) == "0")
+        & ( (df[S47ActualStartDate] < collection_start) | (df[S47ActualStartDate] > reference_date) )
+    )
+    # get all the data that fits the failing condition. Reset the index so that ROW_ID now becomes a column of df
+    df_issues = df[condition].reset_index()
+
+    # SUBMIT ERRORS
+    # Generate a unique ID for each instance of an error. In this case,
+    # - If only LAchildID is used as an identifier, multiple instances of the error on a child will be understood as 1 instance.
+    # We don't want that because in reality, a child can have multiple instances of an error.
+    # - If we use the LAchildID-CPPstartDate combination, that artificially cancels out the instances where a start date repeats for the same child.
+    # Another rule checks for that condition. Not this one.
+    # - It is very unlikely that a combination of LAchildID-CPPstartDate-CPPendDate will repeat in the DataFrame.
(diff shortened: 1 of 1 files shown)


SLornieCYC:
Hi @tab1tha can you help me with what I should do here.  

This is currently failing because I have multiple S47s in-error for the same `LAchildID` with the same values for `S47ActualStartDate` and `ICPCnotRequired` (which can and does happen in real-world data!), which I expected to return as two rows in `assert len(issue_rows)` and `expected_df`.  

However the groupby makes this return as one row with `ROW_ID` = [4, 5].  I'm not sure whether that's allowed?  Can I just change my test output to match what is being returned by the function?

image

There are no other fields I can add to the zipped `link_id` to increase uniqueness of the rows here. `DateOfInitialCPC` won't help because for in-error records this will always be null, and there's no point adding `CINdetailsID` either because both S47s could be in the same CIN episode.

willLPD2I:
Hi @SLornieCYC I've spent a while tinkering with this. Even though in real data someone shouldn't have two S47 start dates on the same date, because your sample data did, the df_issues groupby statement thought they were the same as they had the same link id, as link id is made of LAchildID and s47startdate in your rule. 

There are two fixes to this, I think, one is to add CINdetailsID to the groupby statement so the rule knows that, as the CINdetailsIDs will be different, they're not from the same S47, OR, add CINdetailsID to the link id. The first option works (i've tried) but isn't what you should do because it means issues_df isn't in the format the rest fot he backend wants, so, putting CINdetailsID as a third part of the tuple that make sup link id is best. It works in the code, I've tried it, but there might be something to do with CS data that means this isn't a viable solution?  I'd love to hear your thoughts.

willLPD2I:
As an addendum to that comment, my solution doesn't work if you can have an S47 without a CINdetailsID, or if you can have the same CINdetailsID with different s47s?

SLornieCYC:
Hi @willLPD2I although technically I agree that a child would not normally have multiple S47s on the same day that doesn't mean that combination can't happen in our raw data, which this tool needs to accommodate.

For instance in our CMS if an S47 has two ICPC next actions for the same child (which shouldn't happen but occassionally does - DQ!) the S47 would be returned twice in our CIN S47 data (one row per ICPC).  If neither ICPC had happened at that point the S47 could return in the way I have included in the test (e.g. if the S47 was dated towards the end of the previous year and neither of the two incoming ICPCs had been completed yet).  I admit it's quite an edge case!

In terms of your suggested use of CINdetailsID and your addendum:  
- You can definitely have more than one S47 within the same CINdetailsID (and if they have the same start date I would actually expect that to be the case, at least in our data).
- Depending on how ones reporting is configured to map the referral for CINdetailsID (and the underlying data quality!) it is equally possible to end up with S47s that don't have a CINdetailsID - ditto for content in other module in fact.
- Again depending on ones reporting configuration (this case doesn't apply to us) it might even be possible for the same S47 to be returned against multiple CINdetailsIDs if the referral episode data is wrong (e.g. overlapping episodes).

tab1tha:
Hi @SLornieCYC, @willLPD2I 

Apologies. I was not able to respond sooner. 
This is okay. 

I'll add more comments in the file to explain why the result comes out as [4, 5] combined. There's a line that groups together all rows that have the same ERROR_ID.

cin_validator/rules/cin2022_23/rule_8740.py
```diff
@@ -0,0 +1,205 @@
...
+    # then the <S47ActualStartDate> (N00148) must be on or between [Start_Of_Census_Year] and <ReferenceDate> (N00603)
+    condition = (
+        df[DateOfInitialCPC].isna()
+        & (df[ICPCnotRequired].astype(str) == "0")
+        & (
+            (df[S47ActualStartDate] < collection_start)
+            | (df[S47ActualStartDate] > reference_date)
+        )
+    )
+    # get all the data that fits the failing condition. Reset the index so that ROW_ID now becomes a column of df
```
tab1tha, line 62: Just how it is supposed to be.
This is good work.

cin_validator/rules/cin2022_23/rule_8740.py
```diff
@@ -0,0 +1,205 @@
...
+    # - If we use the LAchildID-CPPstartDate combination, that artificially cancels out the instances where a start date repeats for the same child.
+    # Another rule checks for that condition. Not this one.
+    # - It is very unlikely that a combination of LAchildID-CPPstartDate-CPPendDate will repeat in the DataFrame.
+    # Hence, it can be used as a unique identifier of the row.
+
+    # Replace CPPstartDate and CPPendDate below with the columns concerned in your rule.
+    link_id = tuple(zip(df_issues[LAchildID], df_issues[S47ActualStartDate]))
+    df_issues["ERROR_ID"] = link_id
+    print("\n df_issues with link_id \n", df_issues)
+    df_issues = df_issues.groupby("ERROR_ID")["ROW_ID"].apply(list).reset_index()
```
tab1tha, line 78: Before this line (line 78), df_issues looks exactly like you have made it to be in your expected df.
Line 78 then groups together all the ROW_IDs that have the same ERROR_ID such that df_issues now becomes a dataframe of 

> {
                "ERROR_ID": (
                    "child1",
                    pd.to_datetime("26/05/1999", format="%d/%m/%Y", errors="coerce"),
                ),
                "ROW_ID": [4,5],
            },

Since the ERROR_IDs formed from both rows of your failing data are the same.

cin_validator/rules/cin2022_23/rule_8740.py
```diff
@@ -0,0 +1,205 @@
...
+                "ROW_ID": [4],
+            },
+            {
+                "ERROR_ID": (
+                    "child1",
+                    pd.to_datetime("26/05/1999", format="%d/%m/%Y", errors="coerce"),
+                ),
+                "ROW_ID": [5],
+            },
+        ]
```
tab1tha, line 192: The ROW_IDs are grouped according to their ERROR_IDs so this will be:

> {
"ERROR_ID": (
"child1",
pd.to_datetime("26/05/1999", format="%d/%m/%Y", errors="coerce"),
),
"ROW_ID": [4,5],
},

tab1tha (changes requested):
Just a few finishing touches left. 
We're almost there :)

cin_validator/rules/cin2022_23/rule_8740.py
```diff
@@ -0,0 +1,193 @@
...
+    expected_df = pd.DataFrame(
+        [
+            {
+                "ERROR_ID": (
+                    "child1",
+                    pd.to_datetime("26/05/1999", format="%d/%m/%Y", errors="coerce"),
+                ),
+                "ROW_ID": [4, 5],
+            },
+        ]
```
tab1tha, line 182: It's okay now. 

In the same way, if the child whose data failed on index 4 was different from the child whose data failed on index 5, then the groupby will not group them together and the two separate dataframe rows will be generated as shown:
{
"ERROR_ID": (
"child1",
pd.to_datetime("26/05/1999", format="%d/%m/%Y", errors="coerce"),
),
"ROW_ID": [4,5],
},
{
"ERROR_ID": (
"child2",
pd.to_datetime("26/05/1999", format="%d/%m/%Y", errors="coerce"),
),
"ROW_ID": [5],
},

tab1tha (approved):
Good to go.
Thank you @SLornieCYC

tab1tha:
@amyrichards0n I've added you as a second reviewer. 
We need two approving reviews in order to merge. Could you please give this a look?

tab1tha:
As we said on slack, we'll merge after 24 hours even if a second review is pending.

PR #198, opened 2022-11-18
Rule8720 Issue #101

amyrichards0n:


cin_validator/rules/cin2022_23/rule_8720.py
```diff
@@ -0,0 +1,110 @@
...
+    # the make_census_period function generates the start and end date so that you don't have to do it each time.
+    collection_start, reference_date = make_census_period(ref_date_series)
+
+    # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
+
+    # Where a CPP module is present, <CPPstartDate> (N00105) must be present and on or before the <ReferenceDate> (N00603)
+    failing_indices = df[
+        (df[CPPstartDate] < collection_start) | (df[CPPstartDate] > reference_date)
+    ].index
+
```
willLPD2I, line 52: Hi @amyrichards0n , I'm not quite sure this logic works for this rule. This looks like it's checking if CPPstartDate is within the census period, where 8720 is looking at, if there is a CP plan, the plan must start before the reference date. 

I think the confusion comes from the example linked on the issue page which is for a rule that checks if something is within the census period.

I think your rule wants to check if there's a CPPID, which tells us if there is a CPP module, and if there is, if there's a CP plan start date, and it also wants to check if CP plan start dates are before the reference period.

Does this sound right @tab1tha ?

tab1tha: Having a look ...

tab1tha: Hi @amyrichards0n, the rule logic
> Where a CPP module is present, <CPPstartDate> (N00105) must be present and on or before the <ReferenceDate> (N00603)

can be broken into these two parts. Flag locations where

1. CPP module is present and CPPstartDate is not present. [`CPPID.notna() & CPPstartDate.isna()`]
2. CPPstartDate is present but it is greater than ReferenceDate. [CPPstartDate.notna() & (CPPstartDate > ReferenceDate)]

As Will said, this rule does need you to get the reference date from the Header table. However, since it doesn't require you to check that the date exists within the census period, you don't have to use the make_census_period function. You can go ahead and compare using the ReferenceDate as it is.

Sidenote. You might also notice that the condition `[CPPstartDate.notna() & (CPPstartDate > ReferenceDate)]` can be replaced with `[CPPstartDate > ReferenceDate]` because this check ignores NaNs by default. That is, if either of the dates are missing, the result of the comparison of dates will be NaN and so that position will not be picked by the filter.

Does this help?.

tab1tha:
Hi @amyrichards0n, how is this going?
are you facing any blockers?. 
Sidenote: If you prefer that Will and I complete this one, we can do that too. No pressure.

amyrichards0n:
Hi @tab1tha, sorry I've been struggling for time over the last couple of weeks. Yes, happy for you & Will to complete?

tab1tha:
> Hi @tab1tha, sorry I've been struggling for time over the last couple of weeks. Yes, happy for you & Will to complete?

Sure. We got you !

PR #201, opened 2022-11-22
Rule4013

Csmyth201:
doing a pull request as requested so the code can be reviewed


tab1tha:
Hi @Csmyth201 
You were right, the problem is from the function. 
If you update your branch by running `git merge main`, you'll get a fixed version of the function and your rule will run.
Does it work now?

willLPD2I:
> Hi @Csmyth201 You were right, the problem is from the function. If you update your branch by running `git merge main`, you'll get a fixed version of the function and your rule will run. Does it work now?

If that doesn't work you might need git merge origin/main

tab1tha (approved):

PR #202, opened 2022-11-22
group keys set to false in rules raising warnings

willLPD2I:
fixes warnings in run-all


tab1tha (approved):
Good catch.

Thanks Will !.

tab1tha:
Don't be worried about remembering to format, when you start using git from the command line, running "git commit" will automatically trigger the formatting process.

PR #203, opened 2022-11-22
Rule 8670Q

phoebefryer:


phoebefryer:
@willLPD2I yes, I feel like there should be a better way than I've done but my quick search for an existing function didn't yield anything, but it definitely wasn't the most thorough of googles. I'd be interested in knowing what can do it, or a more efficient loop, I'm sure it could be cut down a little bit

tab1tha:
Found something !

You can replace that loop with this line
`latest_date = collection_end - pd.tseries.offsets.BDay(45)`

BDay makes sure that it only considers business days. It is pandas' business day date offset. @phoebefryer @willLPD2I 

https://stackoverflow.com/questions/44288465/pandas-to-timedelta-with-business-days

willLPD2I:
@phoebefryer , I might be having a silly moment here, but the rule says 'Assessment started more than 45 working days before the end of the census year' and you're failing rows that are within 45 days of the end of the census year, so, less than 45 days before the end of the census year? I think it's checking that assessments don't go longer then 45 working days? I've spent so long reading and re reading the rule though that I think I'm confusing myself and I might have to come to it with fresh eyes.

phoebefryer:
@willLPD2I I think you're right, I think I must have got myself in a bit of a twist with that one with the phrasing. I've now amended it so it should now be right

willLPD2I:
> @willLPD2I I think you're right, I think I must have got myself in a bit of a twist with that one with the phrasing. I've now amended it so it should now be right

Okay! I know when I look at something too long it loses meaning and I spent so long looking at this this morning that I couldn't remember what means what!

cin_validator/rules/cin2022_23/rule_8670Q.py
```diff
@@ -0,0 +1,168 @@
...
+    # Find the reference date - 45
+    latest_date = collection_end - pd.tseries.offsets.BDay(45)
+
+    df_issues = df_assessments[
+        df_assessments[AssessmentActualStartDate] < latest_date
+    ].reset_index()
+
+    link_id = tuple(
+        zip(
+            df_issues[AssessmentAuthorisationDate], df_issues[AssessmentActualStartDate]
```
willLPD2I, line 66: Okay, so if the code checks that assessment authorisation data .isna(), then this makes the first part of this link ID a NaN, so LAchildID might be better!

cin_validator/rules/cin2022_23/rule_8670Q.py
```diff
@@ -0,0 +1,168 @@
...
+            },
+            {
+                "AssessmentAuthorisationDate": pd.NA,
+                "AssessmentActualStartDate": "15/10/2022",
+                # Pass, start date is before ref date - 45wd
+            },
+            {
+                "AssessmentAuthorisationDate": pd.NA,
+                "AssessmentActualStartDate": "15/03/2023",
+                # Fail, start date is withing 45wd of reference date
```
willLPD2I, line 100: Can you update these comments to reflect the rule change please?

willLPD2I (changes requested):

tab1tha:
> > @willLPD2I I think you're right, I think I must have got myself in a bit of a twist with that one with the phrasing. I've now amended it so it should now be right
> 
> Okay! I know when I look at something too long it loses meaning and I spent so long looking at this this morning that I couldn't remember what means what!

same

tab1tha:
@phoebefryer 
Side note: For finishing touches like this, if you prefer that Will and I wrap it up, we'll be happy to do that too.

cin_validator/rules/cin2022_23/rule_8670Q.py
```diff
@@ -0,0 +1,168 @@
...
+                "ERROR_ID": (
+                    "ID2",
+                    pd.to_datetime("15/10/2022", format="%d/%m/%Y", errors="coerce"),
+                ),
+                "ROW_ID": [1],
+            },
+        ]
+    )
+
+    print(issue_rows)
```
tab1tha, line 159: I'll do a pull request to remove these.

PR #204, opened 2022-11-24
rule 8530Q

phoebefryer:


cin_validator/rules/cin2022_23/rule_8530Q.py
```diff
@@ -0,0 +1,167 @@
...
+    #  <ExpectedPersonBirthDate> (N00098) should be between [<ReferenceDate> (N00603) minus 30 days] and [<ReferenceDate> (N00603) plus 9 months]
+
+    # Filter to only those with expected birthdate
+    df = df[~df[ExpectedPersonBirthDate].isna()]
+
+    # Find the reference date - 30
+    earliest_date = collection_start - dt.timedelta(30)
+    # Find reference date + 9 months
+    latest_date = collection_end + dt.timedelta(9 * 30)
+
```
phoebefryer, line 57: If the 9 months needs to be calculated more accurately we will need an extra package installing ( `from dateutil.relativedelta import relativedelta` )

PhilDeakin: @phoebefryer You can use   collection_end + pd.DateOffset(months=9)
It also has days, weeks, years etc....

And it's a part of the pandas module so there's no need to install anything extra.

phoebefryer: Of course pandas has something to do the job! I've amended it now to use that instead

cin_validator/rules/cin2022_23/rule_8530Q.py
```diff
@@ -7,7 +7,6 @@
 from typing import Mapping
 
 import pandas as pd
-import datetime as dt
```
PhilDeakin: Removed as can be done with pd.dateoffset

cin_validator/rules/cin2022_23/rule_8530Q.py
```diff
@@ -51,12 +50,12 @@ def validate(
...
     # Find the reference date - 30
-    earliest_date = collection_start - dt.timedelta(30)
+    earliest_date = collection_start - pd.DateOffset(days=30)
     # Find reference date + 9 months
     latest_date = collection_end + pd.DateOffset(months=9)
-
-    condition1 = df[ExpectedPersonBirthDate] > latest_date
-    condition2 = df[ExpectedPersonBirthDate] < earliest_date
+    
+    condition1 = df[ExpectedPersonBirthDate] >= latest_date
```
PhilDeakin, line 57: Think it's supposed to be inclusive between... I could be wrong.

cin_validator/rules/cin2022_23/rule_8530Q.py
```diff
@@ -155,8 +154,6 @@ def test_validate():
         ]
     )
 
-    print(issue_rows)
-    print(expected_df)
```
PhilDeakin: Removed print statements....

PhilDeakin (approved):
Good stuff

tab1tha:
Hi @PhilDeakin, thank you for this.
@phoebefryer tends to do her finishing changes by herself though. You can make comments on what you think needs to be changed and then tag her.

In general, let's maintain that only Will and Tambe push can push to all branches.

I'll give this a look now :)

cin_validator/rules/cin2022_23/rule_8530Q.py
willLPD2I: The rule looks like it's looking at ReferenceDate, not collection start and end. I think the idea is that if the expected birth date isn't a month before or 9 months after the reference date, either you'd have a birth date, or you wouldn't know you were pregnant to have an expected birth date?

willLPD2I (changes requested):

willLPD2I (approved):

PR #205, opened 2022-11-24
Rule8867

GBurrel:


tab1tha (approved):

cin_validator/rules/cin2022_23/rule_8867.py
```diff
@@ -0,0 +1,243 @@
...
+    condition = df_merged[AssessmentAuthorisationDate].isna()
+    df_merged = df_merged[condition].reset_index()
+
+    # create an identifier for each error instance.
+    # In this case, the rule is checked for each CINclosureDate and AssessmentAuthorisationDate in each child (differentiated by LAchildID)
+    # So, a combination of LAchildID, CINclosureDate and AssessmentAuthorisationDate identifies and error instance.
+    df_merged["ERROR_ID"] = tuple(
+        zip(df_merged[LAchildID], df_merged[CINdetailsID], df_merged[CINclosureDate], df_merged[AssessmentAuthorisationDate])
+    )
+
```
tab1tha, line 78: Since all the failing positions will have the same value of AssessmentAuthorisationDate (that is pd.NA in this case), it doesn't add to the uniqueness of the ERROR_IDs so we can remove it. 

I'll fix the formatting, remove that column and merge it in. 

Thank you @GBurrel,
This is good work.

tab1tha (approved):

cin_validator/rules/cin2022_23/rule_8867.py
```diff
@@ -0,0 +1,243 @@
...
+    result = run_rule(
+        validate,
+        {
+            CINdetails: sample_cind,
+            Assessments: sample_ass,
+        },
+    )
+    
+    # Use .type2_issues to check for the result of .push_type2_issues() which you used above.
+    issues_list = result.type3_issues
```
tab1tha, line 183: oh wait, this should be type2 instead of type3 since you pushed to type2.

tab1tha: Let me run this locally again.

tab1tha:
These are the changes I made:
- Since all the failing positions in this rule will have the same value of AssessmentAuthorisationDate( pd.NA in this case), it doesn't add to the uniqueness of the ERROR_ID so I've removed it.
- The validate function rightly used push_type2 (since merging of tables is involved). However, the test_validate function attempted to check for the pushed value in `result.type3_issues`. I've redirected it to check in the result.type2_issues.
- len(issues_list) verifies the number of tables that were pushed. That is, how many times you called push_type2. In this case, you called it twice at the end of the validate function so two table issue groups will be present in issues list - one for each time you pushed. It is not affected by the number of errors found. You did it correctly, this is additional clarification.
- I noticed that you choose to verify the issue positions flagged in the assessments table (as you went ahead to check that `issues.table == Assessments`). However, to get the Assessments table out of the issues_list, it will be on the second position since it was pushed as the second thing in the validate function. So we'll have to do `issues = issues_list[1]`
- Your ERROR_ID definition (rightly) included CINdetailsID. However, its values weren't included in the expected_df in the test_validate() function. So I've added those in.
- The `assert `issue_rows.equals(expected_df)` wasn't working for a while. I added an alternative way of checking that the dataframes are equal (that is, comparing each column on both sides). It is not necessary since the first assert finally worked. However, I left it in so that you can see another way in which that same check could be implemented.
- The result.definition.message check in the last line was failing. This could be because of a trailing whitespace somewhere. So I copied out the exact string that you used above the validate function and pasted it in that last sentence.

This is ready to go @GBurrel

tab1tha (approved):

PR #206, opened 2022-11-24
Rule8620

GBurrel:


cin_validator/rules/cin2022_23/rule_8620.py
```diff
@@ -0,0 +1,111 @@
...
+CINclosureDate = CINdetails.CINclosureDate
+Header = CINTable.Header
+ReferenceDate = Header.ReferenceDate
+
+# define characteristics of rule
+@rule_definition(
+    code=8620,
+    module=CINTable.CINdetails,
+    message="CIN Closure Date present and does not fall within the Census year",
+    affected_fields=[CINclosureDate,ReferenceDate],
```
willLPD2I, line 27: So, reference date is relevant to the rule, but we don't count it as an affected field because this rule isn't really checking if it's wrong.

cin_validator/rules/cin2022_23/rule_8620.py
```diff
@@ -0,0 +1,111 @@
+from typing import Mapping
```
willLPD2I, line 1: if you could format this by running 'black /workspaces/CIN-validator/cin_validator/rules/cin2022_23' in the terminal, Gavin, that'd be really good so we don't get the linting error. I tend to use the filepath of the whole cin_validator module rather than just the rule file as it'll fix any other files with issues too.

cin_validator/rules/cin2022_23/rule_8620.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
+
+    # If <CINclosureDate> (N00102) is present, it must be within [Period_of_Census]
+    df = df[
+        ~(
+            (df[CINclosureDate] > collection_start)
+            & (df[CINclosureDate] < collection_end)
+        )
+    ]
```
willLPD2I, line 53: Looks good!

willLPD2I (changes requested):
This all looks good Gavin, if you can just amend the affected fields line and run black to format, I'll give it a final check and I think we're good to go

GBurrel:
Hi @willLPD2I  - I've made the required change to thew code and also ran the formatting commanding. Hopefully the saved changes have now come through.

willLPD2I (approved):

GBurrel:
Hi @willLPD2I - is this one complete now?

tab1tha:
Hi @GBurrel, I have linked this pull request to the issue it addresses by clicking on the cog beside the "Developement" tag in the right hand side, and searching for the rule code.
image

tab1tha (approved):

PR #207, opened 2022-11-25
Added rule 4016

### LINKED ISSUE #85: Rule 4016 : A CIN Plan has been reported as open at the same time as a Child Protection Plan.
The \<CINPlanStartDate\> (N00689) for any CIN Plan group cannot fall within either:  \<CPPstartDate\> (N00105) or \<CPPendDate\> (N00115); or \<CPPstartDate\> and \<ReferenceDate\> (N00603) if \<CPPendDate\> is not present - for any CPP group; unless \<CINPlanStartDate\> is equal to \<CPPendDate\> for this group

SLornieCYC:
Closes #85

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4016.py
+++ b/cin_validator/rules/cin2022_23/rule_4016.py
@@ -79,13 +79,19 @@ def validate(
         suffixes=("_cin", "_cpp"),
     )
 
-    #  Get rows where CINPlanStartDate is between CPPstartDate and CPPendDate (or reference_date)
-    condition = (df_merged[CINPlanStartDate] >= df_merged[CPPstartDate]) & (
+    # Get rows where CINPlanStartDate is after CPPstartDate 
+    # and CINPlanStartDate before CPPendDate (or if null, before/on ReferenceDate)
+    cin_start_after_cpp_start = (df_merged[CINPlanStartDate] >= df_merged[CPPstartDate])
+    cin_start_before_cpp_end = (
         (df_merged[CINPlanStartDate] < df_merged[CPPendDate])
         & df_merged[CPPendDate].notna()
-        | (df_merged[CINPlanStartDate] < reference_date) & df_merged[CPPendDate].isna()
-    )
-    df_merged = df_merged[condition].reset_index()
+        )
+    cin_start_before_reference_date = (
+        (df_merged[CINPlanStartDate] <= reference_date) 
+        & df_merged[CPPendDate].isna()
+        )
+    
+    df_merged = df_merged[cin_start_after_cpp_start & (cin_start_before_cpp_end | cin_start_before_reference_date)].reset_index()
 
     # create an identifier for each error instance.
     df_merged["ERROR_ID"] = tuple(
@@ -156,6 +162,12 @@ def test_validate():
                 "CPPendDate": "30/10/2001",
                 "CPPID": "cinID1",
             },
+            {
+                "LAchildID": "child5",
+                "CPPstartDate": "26/05/2000",
+                "CPPendDate": pd.NA,
+                "CPPID": "cinID1",
+            },
         ]
     )
     sample_cin = pd.DataFrame(
@@ -199,6 +211,10 @@ def test_validate():
             {
                 "LAchildID": "child4",  # 9 Pass - No CPP
                 "CINPlanStartDate": "04/06/2000",
+            },
+            {
+                "LAchildID": "child5",  # 10 Fail - Start on ReferenceDate
+                "CINPlanStartDate": "31/03/2001",
             },
         ]
     )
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_4016.py
```diff
@@ -0,0 +1,306 @@
...
+        on=["LAchildID"],
+        how="left",
+        suffixes=("_cin", "_cpp"),
+    )
+
+    #  Get rows where CINPlanStartDate is between CPPstartDate and CPPendDate (or reference_date)
+    condition = (df_merged[CINPlanStartDate] >= df_merged[CPPstartDate]) & (
+        (df_merged[CINPlanStartDate] < df_merged[CPPendDate])
+        & df_merged[CPPendDate].notna()
+        | (df_merged[CINPlanStartDate] < reference_date) & df_merged[CPPendDate].isna()
```
willLPD2I, line 86: You remember in maths at school they always used to say, do one bit of the sum at a time so you can go back and see where you went wrong later, this feels a bit like that. It's taking me a little while to parse mentally.

willLPD2I: @SLornieCYC, I'm not entirely sure this works quite right, I think we need to separate some logical conditions with brackets a bit more to make sure it's doing what we think, or, we need to write them as separate conditions and combine them in the line that does the slicing. Even if it does work, the readability means that people who take over or help maintain the code in the future will struggle so we should refactor.

At the moment it checks if (CPPstart is between CINstart and CINend AND CINend is not NaN) or (CINstart is less than reference date) and (CPPend is NaN)

What we want to check is:
The <CINPlanStartDate> (N00689) for any CIN Plan group cannot fall within either:
<CPPstartDate> (N00105) or <CPPendDate> (N00115);
or <CPPstartDate> and <ReferenceDate> (N00603) if <CPPendDate> is not present - for any CPP group;
unless <CINPlanStartDate> is equal to <CPPendDate> for this group

Which is:
Fail if CPPstart < CINstart < CPPend
or fail if CPPstart < CINstart < RefDate if CPPend is not present
But allow both of these to pass is CINstart = CPPend for a group

What I think wouyld be better, especially as far as readability is something like 
condition 1 = CPPstart is between CINstart and CINend 
condition 2 =  CPPend isn't present and CPPstart is between CINstart and Reference date
condition 3 = CINstart is the same as CPPend in a group

Then we combine with something like: (condition 1 or condition 2) or condition 3

cin_validator/rules/cin2022_23/rule_4016.py
```diff
@@ -0,0 +1,306 @@
...
+
+    # lOGIC
+    # Implement rule logic as described by the Github issue.
+    # Put the description as a comment above the implementation as shown.
+
+    # The <CINPlanStartDate> (N00689) for any CIN Plan group cannot fall within either:
+    # <CPPstartDate> (N00105) or <CPPendDate> (N00115);
+    # or <CPPstartDate> and <ReferenceDate> (N00603) if <CPPendDate> is not present - for any CPP group;
+    # unless <CINPlanStartDate> is equal to <CPPendDate> for this group
+
```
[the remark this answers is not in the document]
tab1tha: Nope, on second thought don't do that here because the condition has an exception only for CPPendDate and not reference_date. So we'll need to know which one we're comparing CINplanStartDate with.

I'm referring to this clause: "unless CINPlanStartDate is equal to CPPendDate for this group"

cin_validator/rules/cin2022_23/rule_4016.py
```diff
@@ -0,0 +1,306 @@
...
+    condition = (df_merged[CINPlanStartDate] >= df_merged[CPPstartDate]) & (
+        (df_merged[CINPlanStartDate] < df_merged[CPPendDate])
+        & df_merged[CPPendDate].notna()
+        | (df_merged[CINPlanStartDate] < reference_date) & df_merged[CPPendDate].isna()
+    )
+    df_merged = df_merged[condition].reset_index()
+
+    # create an identifier for each error instance.
+    df_merged["ERROR_ID"] = tuple(
+        zip(df_merged[LAchildID], df_merged[CINPlanStartDate])
```
tab1tha, line 92: Good choice of link id.

tab1tha (changes requested):
This is good work @SLornieCYC 
Will and I have left some comments in our review.

It might need a tweak or two.

cin_validator/rules/cin2022_23/rule_4016.py
SLornieCYC: Thanks Will. Have now split up the condition logic to make it more readable. Don't think what I have is incorrect though.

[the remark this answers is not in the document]
SLornieCYC: Thanks. Have split the conditions part and updated < to <= for reference date.

willLPD2I (approved):

tab1tha (approved):

PR #209, opened 2022-11-25
Added rule 8940

### LINKED ISSUE #110: Rule 8940 : Child Protection Plan data contains overlapping dates
Where more than one \<ChildProtectionPlans\> group is included, the \<CPPstartDate\> (N00105) of each group cannot fall within either:

a) \<CPPstartDate\> (N00105) to \<CPPendDate\> (N00115), or

b) \<CPPstartDate\> (N00105) and \<ReferenceDate\> if \<CPPendDate\> (N00115) is not present

of any other group

SLornieCYC:
Closes #110

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8940.py
+++ b/cin_validator/rules/cin2022_23/rule_8940.py
@@ -123,10 +123,10 @@ def validate(
     print("\n df_cpp2_issues\n", df_cpp2_issues)
 
     # Ensure that you maintain the ROW_ID, and ERROR_ID column names which are shown above. They are keywords in this project.
-    rule_context.push_type_2(
+    rule_context.push_type_3(
         table=ChildProtectionPlans, columns=[CPPstartDate], row_df=df_cpp_issues
     )
-    rule_context.push_type_2(
+    rule_context.push_type_3(
         table=ChildProtectionPlans,
         columns=[CPPstartDate, CPPendDate],
         row_df=df_cpp2_issues,
@@ -213,7 +213,7 @@ def test_validate():
     )
 
     # Use .type2_issues to check for the result of .push_type2_issues() which you used above.
-    issues_list = result.type2_issues
+    issues_list = result.type3_issues
     assert len(issues_list) == 2
     # the function returns a list on NamedTuples where each NamedTuple contains (table, column_list, df_issues)
     # pick any table and check it's values. the tuple in location 1 will contain the Reviews columns because that's the second thing pushed above.
--- a/cin_validator/rules/cin2022_23/rule_8940.py
+++ b/cin_validator/rules/cin2022_23/rule_8940.py
@@ -72,7 +72,7 @@ def validate(
     #  Create dataframes which only have rows with CP plans, and which should have one plan per row.
     df_cpp = df_cpp[df_cpp[CPPstartDate].notna()]
     df_cpp2 = df_cpp2[df_cpp2[CPPstartDate].notna()]
-    print("\n df_cpp\n", df_cpp)
+
     #  Merge tables to test for overlaps
     df_merged = df_cpp.merge(
         df_cpp2,
@@ -83,7 +83,7 @@ def validate(
 
     # Exclude rows where the CPPID is the same on both sides
     df_merged = df_merged[(df_merged["CPPID_cpp"] != df_merged["CPPID_cpp2"])]
-    print("\n df_merged exclude\n", df_merged)
+
     # Determine whether CPP overlaps another CPP
     cpp_started_after_start = (
         df_merged["CPPstartDate_cpp"] >= df_merged["CPPstartDate_cpp2"]
@@ -98,13 +98,13 @@ def validate(
     df_merged = df_merged[
         cpp_started_after_start & (cpp_started_before_end | cpp_started_before_refdate)
     ].reset_index()
-    print("\n df_merged filtered\n", df_merged)
+
     # create an identifier for each error instance.
     # In this case, the rule is checked for each CPPstartDate, in each CPplanDates group (differentiated by CP dates), in each child (differentiated by LAchildID)
     df_merged["ERROR_ID"] = tuple(
         zip(df_merged[LAchildID], df_merged["CPPID_cpp"], df_merged["CPPID_cpp2"])
     )
-    print("\n df_merged zipped\n", df_merged)
+
     # The merges were done on copies of cpp_df so that the column names in dataframes themselves aren't affected by the suffixes.
     # we can now map the suffixes columns to their corresponding source tables such that the failing ROW_IDs and ERROR_IDs exist per table.
     df_cpp_issues = (
@@ -119,8 +119,6 @@ def validate(
         .apply(list)
         .reset_index()
     )
-    print("\n df_cpp_issues\n", df_cpp_issues)
-    print("\n df_cpp2_issues\n", df_cpp2_issues)
 
     # Ensure that you maintain the ROW_ID, and ERROR_ID column names which are shown above. They are keywords in this project.
     rule_context.push_type_3(
@@ -231,7 +229,7 @@ def test_validate():
     issue_rows = issues.row_df
     # replace 3 with the number of failing points you expect from the sample data.
     assert len(issue_rows) == 3
-    print("\n issue_rows\n", issue_rows)
+
     # check that the failing locations are contained in a DataFrame having the appropriate columns. These lines do not change.
     assert isinstance(issue_rows, pd.DataFrame)
     assert issue_rows.columns.to_list() == ["ERROR_ID", "ROW_ID"]
--- a/cin_validator/rules/cin2022_23/rule_8940.py
+++ b/cin_validator/rules/cin2022_23/rule_8940.py
@@ -89,10 +89,10 @@ def validate(
         df_merged["CPPstartDate_cpp"] >= df_merged["CPPstartDate_cpp2"]
     )
     cpp_started_before_end = (
-        df_merged["CPPstartDate_cpp"] < df_merged["CPPendDate_cpp2"]
+        df_merged["CPPstartDate_cpp"] <= df_merged["CPPendDate_cpp2"]
     ) & df_merged["CPPendDate_cpp2"].notna()
     cpp_started_before_refdate = (
-        df_merged["CPPstartDate_cpp"] < reference_date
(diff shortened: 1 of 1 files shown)


SLornieCYC:
**Note**: I've coded this as a type 2 rule because I'm joining ChildProtectionPlans back onto itself, whereas type 3 seems to be for grouped data and I couldn't work out how this fit that structure.  Hopefully that's okay?

willLPD2I:
> **Note**: I've coded this as a type 2 rule because I'm joining ChildProtectionPlans back onto itself, whereas type 3 seems to be for grouped data and I couldn't work out how this fit that structure. Hopefully that's okay?
> 
> Print statements included in initial commit for testing purposes. Will remove later.

Hi @SLornieCYC , so this rule can actually be written as a Type 3 rule rather than a Type 2, @tab1tha has written a guide here explaining how to do it: https://github.com/data-to-insight/CIN-validator/blob/main/Documentation/rule_logic_demos/type3_compare_intervals_demo.ipynb

Now, your version seems to work so doesn't need to be refactored in terms of the logic, as far as I can see, however, I'm not sure if it needs to be changed so that the Type 3 context can be used so that the correct information is collected to be sent to display correctly on the front end, @tab1tha do you have any ideas about that?

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+    df_cpp2 = df_cpp2[df_cpp2[CPPstartDate].notna()]
+    print("\n df_cpp\n", df_cpp)
+    #  Merge tables to test for overlaps
+    df_merged = df_cpp.merge(
+        df_cpp2,
+        on=["LAchildID"],
+        how="left",
+        suffixes=("_cpp", "_cpp2"),
+    )
+
```
tab1tha, line 83: Nice

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+    cpp_started_after_start = (
+        df_merged["CPPstartDate_cpp"] >= df_merged["CPPstartDate_cpp2"]
+    )
+    cpp_started_before_end = (
+        df_merged["CPPstartDate_cpp"] < df_merged["CPPendDate_cpp2"]
+    ) & df_merged["CPPendDate_cpp2"].notna()
+    cpp_started_before_refdate = (
+        df_merged["CPPstartDate_cpp"] < reference_date
+    ) & df_merged["CPPendDate_cpp2"].isna()
+
```
tab1tha, line 97: Good initiative. I can see your thought process.

One other way of doing this could be to replace all the NaN values in CPPendDate of cpp2 before merging.  You could do that by
`df_cpp2[CppEndDate] = df_cpp2[CppEndDate].fillna(reference_date)`

tab1tha: You don't have to change how you've done it. 
Just mentioning.

You can read about fillna [here in the docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html) and [here on stackoverflow](https://stackoverflow.com/questions/38134012/pandas-dataframe-fillna-only-some-columns-in-place)

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+    # Ensure that you maintain the ROW_ID, and ERROR_ID column names which are shown above. They are keywords in this project.
+    rule_context.push_type_2(
+        table=ChildProtectionPlans, columns=[CPPstartDate], row_df=df_cpp_issues
+    )
+    rule_context.push_type_2(
+        table=ChildProtectionPlans,
+        columns=[CPPstartDate, CPPendDate],
+        row_df=df_cpp2_issues,
+    )
```
tab1tha, line 134: This is one of the scenarios where both push_type2 and push_type3 work. 

Still, if you replace these with push_type3, then the rule will be classified properly as a type3 rule in case we want to filter later one. Could you do that?.

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+            },
+            {
+                "LAchildID": "child3",  # 5 Fail
+                "CPPstartDate": "26/08/2000",
+                "CPPendDate": "26/10/2000",
+                "CPPID": "cinID32",
+            },
+        ]
+    )
+
```
tab1tha, line 183: Niceeee
Full test coverage !

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+    result = run_rule(
+        validate,
+        {
+            ChildProtectionPlans: sample_cpp,
+            Header: sample_header,
+        },
+    )
+
+    # Use .type2_issues to check for the result of .push_type2_issues() which you used above.
+    issues_list = result.type2_issues
```
tab1tha, line 205: Do replace type2 by type3 here too.

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+    df_merged = df_merged[
+        cpp_started_after_start & (cpp_started_before_end | cpp_started_before_refdate)
+    ].reset_index()
+    print("\n df_merged filtered\n", df_merged)
+    # create an identifier for each error instance.
+    # In this case, the rule is checked for each CPPstartDate, in each CPplanDates group (differentiated by CP dates), in each child (differentiated by LAchildID)
+    # So, a combination of LAchildID, CPPstartDate and CPPreviewDate identifies and error instance.
+    df_merged["ERROR_ID"] = tuple(
+        zip(df_merged[LAchildID], df_merged["CPPID_cpp"], df_merged["CPPID_cpp2"])
+    )
```
tab1tha, line 107: Just the way it should be !

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+        .groupby("ERROR_ID", group_keys=False)["ROW_ID"]
+        .apply(list)
+        .reset_index()
+    )
+    df_cpp2_issues = (
+        df_cpp2.merge(df_merged, left_on="ROW_ID", right_on="ROW_ID_cpp2")
+        .groupby("ERROR_ID", group_keys=False)["ROW_ID"]
+        .apply(list)
+        .reset_index()
+    )
```
tab1tha, line 122: Exactly

tab1tha (changes requested):
Very impressive @SLornieCYC 
very impressive.

You implemented the concepts from the demo well.

tab1tha:
> **Note**: I've coded this as a type 2 rule because I'm joining ChildProtectionPlans back onto itself, whereas type 3 seems to be for grouped data and I couldn't work out how this fit that structure. Hopefully that's okay?
> 
> Print statements included in initial commit for testing purposes. Will remove later.

Hi @SLornieCYC , this is really good work. 
You can change it to use type3 too. push_type3 also accomodates this strucure as tested here in [test_type3](https://github.com/data-to-insight/CIN-validator/blob/c6910822eabbb10e5efb1eada277b5f4c6c0915e/tests/rule_engine/test_context.py#L170).

When you change the type2 phrases to type3 and remove your print statements, you're good to go.

cin_validator/rules/cin2022_23/rule_8940.py
```diff
@@ -0,0 +1,262 @@
...
+        how="left",
+        suffixes=("_cpp", "_cpp2"),
+    )
+
+    # Exclude rows where the CPPID is the same on both sides
+    df_merged = df_merged[(df_merged["CPPID_cpp"] != df_merged["CPPID_cpp2"])]
+    print("\n df_merged exclude\n", df_merged)
+    # Determine whether CPP overlaps another CPP
+    cpp_started_after_start = (
+        df_merged["CPPstartDate_cpp"] >= df_merged["CPPstartDate_cpp2"]
```
tab1tha, line 89: We will have to sync on what "within" means. 
Should the comparisons for start and end be "< and >" or "<= and >="
@SLornieCYC

tab1tha (changes requested):

cin_validator/rules/cin2022_23/rule_8940.py
tab1tha: In the rules where we checked whether a date existed within the period of census, we used "<= and >=".

Here, you've used >= to compare the start date and < to compare the end date.

tab1tha: The description [of this other rule](https://github.com/data-to-insight/CIN-validator/issues/85#issue-1306770976) seems to imply that if a date is equal to the border date then it is within the boundary.
Otherwise, the fact that <CINPlanStartDate> is equal to <CPPendDate> will not need to be stated as a separate condition in that rule.

cin_validator/rules/cin2022_23/rule_8940.py
SLornieCYC: The one thing I'm not sure about is, should type 2 rules test ALL outputs? 

E.g. this returns errors in two different tables but the tests only make sure that one of the tables has output as expected (as in the base rule we are advised to copy).

assert issue_table
assert issue_columns
assert issue_rows

cin_validator/rules/cin2022_23/rule_8940.py
SLornieCYC: Fair point. Have changed both of these to use `<=`.

cin_validator/rules/cin2022_23/rule_8940.py
SLornieCYC: Done!

cin_validator/rules/cin2022_23/rule_8940.py
tab1tha: Yes. 
Ideally, we would check issue_table, issue_columns, and issue_rows for all the tables created in the push. 

Choosing one to check is good but checking all is better (though longer).

tab1tha (approved):

tab1tha (approved):

tab1tha:
@PhilDeakin have you given this a look already?

willLPD2I (approved):

Issue #69, opened 2022-07-16
Rule 8863 : An Assessment is shown as starting when there is another Assessment ongoing 


tab1tha:
Within one \<CINdetails\> group, where present each \<AssessmentActualStartDate\> (N00159) must not fall on or between:

a) the \<AssessmentActualStartDate\> (N00159) and \<AssessmentAuthorisation Date (N00160) of any other \<Assessments\> group that has an \<AssessmentAuthorisationDate\> (N00160)

or

b) the \<AssessmentActualStartDate\> (N00159) and the \<ReferenceDate\> (N00603) where the \<AssessmentAuthorisationDate\> (N00160) is missing

Issue #58, opened 2022-07-16
Rule 8890 : A Section 47 enquiry is shown as starting when there is another Section 47 Enquiry ongoing


tab1tha:
Within one \<CINdetails\> group, each \<S47ActualStartDate\> (N00148) must not fall on or between a) the \<S47ActualStartDate\> (N00148) and \<DateOfInitialCPC\> (N00110) of any other \<Section47\> group that has a \<DateOfInitialCPC\> (N00110) or b) the  \<S47ActualStartDate\> (N00148) and the \<ReferenceDate\> (N00603) of any other \<Section47\> group that has a missing \<DateOfInitialCPC\> (N00110) and the \<ICPCnotRequired\> (N00111) flag is not true

PR #211, opened 2022-11-28
Code for review Closes Rule #69

GBurrel:
#69

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8863.py
+++ b/cin_validator/rules/cin2022_23/rule_8863.py
@@ -1,3 +1,5 @@
+# this is similar to rule 8890
+
 from typing import Mapping
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


willLPD2I:
Hi Stephen, I've made a mistake in my Type 3 rule guide. What you've done perfectly fits that guide, but I didn't realise there is a second type of Type 3 rule. The kind in my guide, and the kind you have written just checks if there are too many open plans. Sadly, this specific rule checks something subtly different: has any plan been open whilst another was still open. 

Luckily for me, Tambe has written a guide on how the logic of this type of rule works: https://github.com/data-to-insight/CIN-validator/blob/main/Documentation/rule_logic_demos/type3_compare_intervals_demo.ipynb However, the way the guide is written is quite abstract. If you want to get in a call to talk about it, do let me know. Alternatively, I'll be going through this type of rule in Thursday's drop-ins, and if I remember rightly, when you can make it, you come on Thursdays.

GBurrel:
Hi @willLPD2I - I assume this message is intended for me? Either way I can confirm I will be at the session this coming Thursday.

tab1tha:
Hi @GBurrel, thank you for this.

The edits that I have done here are similar to those done in this rule https://github.com/data-to-insight/CIN-validator/issues/58. Both rules have similar logic.

PR #214, opened 2022-11-29
Resolve erratic behaviour in run-all command

tab1tha:
The changes in this pull request
- Ensures that each rule runs on a copy of the data to prevent the original data from being changed within the rules.

Source:
It was noticed that the tools produced significantly different results when run locally and when run on codespaces. This turned out to be because of two main reasons,
- some rules were editing the data files (they aren't supposed to)
- the rules were not being run in the same order (local env is windows and codespaces env is linux)

Solution:
- Make a copy of the data each time a rule is run.


willLPD2I (approved):

phoebefryer (approved):

tab1tha:
Thank you !

PR #217, opened 2022-11-30
draft Rules df

willLPD2I:
@tab1tha is this the kinda thing you mean?

### DIFF
--- a/cin_validator/cin_validator_class.py
+++ b/cin_validator/cin_validator_class.py
@@ -21,6 +21,7 @@ class CinValidationSession:
         self.create_error_report_df()
         self.create_json_report()
         self.select_by_id()
+        self.create_rules_df()
 
     def create_error_report_df(self):
         self.issue_instances = pd.DataFrame()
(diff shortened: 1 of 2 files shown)


cin_validator/cin_validator_class.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def create_rules_df(self):
+        # rules_broken = self.all_rules_issue_locs
+        rules_broken = (
+            self.all_rules_issue_locs["rule_code"].astype("str").unique().tolist()
+        )
+        print(rules_broken)
+        rule_messages = []
+        for rule in registry:
+            if str(rule.code) in rules_broken:
```
tab1tha, line 103: Hi Will, this works does work.

On second thought however, let us not loop through registry twice (we are doing that in the create_error_report_df method).
How about merging this feature with the create_error_report_df function such that when we identify the rules that fail, we immediately add the rule code-description pair to a  dictionary as you did with issue_dict?

What do you think? @willLPD2I

tab1tha (changes requested):
Sidenote: I've merged in the parent branch so if you do `git merge main` on this branch, it'll filter out only the changes you've made.

cin_validator/cin_validator_class.py
willLPD2I: @tab1tha Sounds way better! Cool!

tab1tha (approved):
This is good work Will.

tab1tha:
Resolving the merge conflict introduced formatting differences so I'll run formatting now.

PR #219, opened 2022-12-05
Added rule 4017 (mirror of 4016)

### LINKED ISSUE #86: Rule 4017 : A CIN Plan has been reported as open at the same time as a Child Protection Plan.
The \<CPPstartDate\> (N00105) for any CPP group cannot fall within either: \<CINPlanStart Date\> (N00689) or \<CINPlanEndDate\> (N00690); or \<CINPlanStartDate\> and \<ReferenceDate\> (N00603) if \<CINPlanEndDate\> is not present - for any CIN Plan Group; unless \<CCPstartDate\> is equal to \<CINPlanEndDate\> for this group.

SLornieCYC:
Closes #86

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4017.py
+++ b/cin_validator/rules/cin2022_23/rule_4017.py
@@ -30,7 +30,7 @@ ReferenceDate = Header.ReferenceDate
     # Note that even if multiple tables are involved, one table will be named in the module column.
     module=CINTable.ChildProtectionPlans,
     # replace the message with the corresponding value for this rule, gotten from the excel sheet.
-    message="A CIN Plan has been reported as open at the same time as a Child Protection Plan.  ",
+    message="A CIN Plan has been reported as open at the same time as a Child Protection Plan.",
     # The column names tend to be the words within the < > signs in the github issue description.
     affected_fields=[
         CINPlanStartDate,
--- a/cin_validator/rules/cin2022_23/rule_4017.py
+++ b/cin_validator/rules/cin2022_23/rule_4017.py
@@ -28,7 +28,7 @@ ReferenceDate = Header.ReferenceDate
     code=4017,
     # replace ChildProtectionPlans with the value in the module column of the excel sheet corresponding to this rule .
     # Note that even if multiple tables are involved, one table will be named in the module column.
-    module=CINTable.ChildProtectionPlans,
+    module=CINTable.CINplanDates,
     # replace the message with the corresponding value for this rule, gotten from the excel sheet.
     message="A CIN Plan has been reported as open at the same time as a Child Protection Plan.",
     # The column names tend to be the words within the < > signs in the github issue description.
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_4017.py
```diff
@@ -0,0 +1,325 @@
...
+ReferenceDate = Header.ReferenceDate
+
+
+# define characteristics of rule
+@rule_definition(
+    # write the rule code here
+    code=4017,
+    # replace ChildProtectionPlans with the value in the module column of the excel sheet corresponding to this rule .
+    # Note that even if multiple tables are involved, one table will be named in the module column.
+    module=CINTable.ChildProtectionPlans,
```
GBurrel, line 31: Would you not detail the CIN Plan Dates module here rather than Child Protection Plans?

cin_validator/rules/cin2022_23/rule_4017.py
```diff
@@ -0,0 +1,325 @@
...
+    # <CPPstartDate> (N00105) or <CPPendDate> (N00115);
+    # or <CPPstartDate> and <ReferenceDate> (N00603) if <CPPendDate> is not present - for any CPP group;
+    # unless <CINPlanStartDate> is equal to <CPPendDate> for this group
+
+    #  Merge tables
+    df_merged = df_cpp.merge(
+        df_cin,
+        on=["LAchildID"],
+        how="left",
+        suffixes=("_cpp", "_cin"),
```
GBurrel, line 80: Might you also use CINdetailsID to assist with the merging of the tables?

cin_validator/rules/cin2022_23/rule_4017.py
SLornieCYC: There are two rules that share the same description and look at the same data from opposite angles.  For this one (4017 #86) I've selected Child Protection Plans as the module because the rule is about CPP start date within a CIN Plan.  When I wrote the other side before (4016 #85, PR #207) I selected CIN Plan Dates as the module because that rule is about CIN start date within a CPP.

*I have just noticed though that I forgot to update the rule logic at line 70 here when I copied this from my 4016 branch; will correct that later!

cin_validator/rules/cin2022_23/rule_4017.py
SLornieCYC: I could use CINdetailsID as well but unlike some other rules the logic here doesn't mention "within a CINdetailsID group" so I've erred on safe side with a more flexible interpretation.  

E.g. if you have a CPP that starts within a CIN Plan period and _somehow_ they have different CINdetailsIDs in the source data.  Not merging on CINdetailsID here means that overlap will still flag up as an error.

willLPD2I: This seems like the right step to me @SLornieCYC for this particular case.

willLPD2I (approved):

GBurrel:
Thanks for your comments @SLornieCYC  - very helpful to me (as well as the code you wrote).

tab1tha (approved):
This is good to go. 

I'll push the change for the name of the module so that we can wrap it up.

PR #220, opened 2022-12-05
Added rule 8839

### LINKED ISSUE #57: Rule 8839 : Within one CINDetails group there are 2 or more open S47 Assessments
Within one \<CINdetails\> group, there must be only one \<Section47\> group that has no \<DateOfInitialCPC\> (N00110) recorded OR \<Section47\> group that has a missing \<DateOfInitialCPC\> (N00110) and the \<ICPCnotRequired\> (N00111) flag is not true

SLornieCYC:
Closes #57

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8839.py
+++ b/cin_validator/rules/cin2022_23/rule_8839.py
@@ -1,0 +1,219 @@
+from typing import Mapping
+
+import pandas as pd
+
+from cin_validator.rule_engine import CINTable, RuleContext, rule_definition
+from cin_validator.rules.cin2022_23.rule_8925 import LAchildID
+from cin_validator.test_engine import run_rule
+
+# Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
+
+Section47 = CINTable.Section47
+LAchildID = Section47.LAchildID
+CINdetailsID = Section47.CINdetailsID
+DateOfInitialCPC = Section47.DateOfInitialCPC
+ICPCnotRequired = Section47.ICPCnotRequired
+
+# define characteristics of rule
+@rule_definition(
+    # write the rule code here, in place of 8896
+    code=8839,
+    # replace Assessments with the value in the module column of the excel sheet corresponding to this rule .
+    module=CINTable.Section47,
+    # replace the message with the corresponding value for this rule, gotten from the excel sheet.
+    message="Within one CINDetails group there are 2 or more open S47 Assessments",
+    # The column names tend to be the words within the < > signs in the github issue description.
+    affected_fields=[DateOfInitialCPC],
+)
+def validate(
+    data_container: Mapping[CINTable, pd.DataFrame], rule_context: RuleContext
+):
+    # PREPARING DATA
+
+    # Replace Assessments with the name of the table you need.
+    df = data_container[Section47]
+    # Before you begin, rename the index and make it a column, so that the initial row positions can be kept intact.
+    df.index.name = "ROW_ID"
+    df.reset_index(inplace=True)
+
+    # lOGIC
+    # Implement rule logic as described by the Github issue.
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8839.py
```diff
@@ -0,0 +1,214 @@
...
+    # resetting the index pushes the columns in the () back as columns of the dataframe and assigns the groupby result to the column in the square bracket.
+    #
+    # filter out the instances where DateOfInitialCPC is missing more than once in a CINdetails group.
+    df_check = df_check[df_check["CountICPC"] > 1]
+    issue_ids = tuple(
+        zip(df_check[LAchildID], df_check[CINdetailsID], df_check[DateOfInitialCPC])
+    )
+    # DF_ISSUES: GET ALL THE DATA ABOUT THE LOCATIONS THAT WERE IDENTIFIED IN DF_CHECK
+    df["ERROR_ID"] = tuple(zip(df[LAchildID], df[CINdetailsID], df[DateOfInitialCPC]))
+    df_issues = df[df.ERROR_ID.isin(issue_ids)]
```
SLornieCYC, line 78: Note I've deviated from the example in #172 because I found it doesn't work when there are a mixture of open and closed records (Section 47s here, Assessments in 8896) within the same CINdetailsID group.  I have to include the null date in the error ID to avoid flagging those incorrectly in the error table.

cin_validator/rules/cin2022_23/rule_8839.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # Within one <CINdetails> group, there must be only one <Section47> group that has no <DateOfInitialCPC> (N00110) recorded
+    # OR <Section47> group that has a missing <DateOfInitialCPC> (N00110) and the <ICPCnotRequired> (N00111) flag is not true
+
+    # DF_CHECK: APPLY GROUPBYs IN A SEPARATE DATAFRAME SO THAT OTHER COLUMNS ARE NOT LOST OR CORRUPTED. THEN, MAP THE RESULTS TO THE INITIAL DATAFRAME.
+    df_check = df.copy()
+    # get all the locations where ICPCnotRequired is null or not true (1)
+    df_check = df_check[
+        df_check[ICPCnotRequired].isna() | (df_check[ICPCnotRequired] != 1)
+    ]
```
willLPD2I, line 51: I waas thinking about this a good while, it seems right to do the slice like this according to the rule logic, the rule logic is just written less than clearly.

willLPD2I:
@tab1tha this looks okay to me, but I'd like a second opinion for saftey!

willLPD2I (approved):

cin_validator/rules/cin2022_23/rule_8839.py
tab1tha: So for cases where it is necessary to map issue_ids from a filtered df (df_check in this case) to the original dataframe (df in this case), we can say that it is okay to leave in the redundant columns of ERROR_ID since they might act as a differentiating factor during the df matching.

cin_validator/rules/cin2022_23/rule_8839.py
tab1tha: This rule description is not quite clear. 
It is also possible to interpret the first part as "flag the groups which have more than one Section47 whose DateOfInitialCPC is missing even if the ICPCnotRequired is true".

Let's take consider the logic in the way you're interpreted it @SLornieCYC  and see how it works when compared with the DfE tool.

tab1tha:
This rule description is not quite clear.
It is also possible to interpret the first part as "flag the groups which have more than one Section47 whose DateOfInitialCPC is missing even if the ICPCnotRequired is true".

Let's take consider the logic in the way you're interpreted it @SLornieCYC  and see how it works when compared with the DfE tool. @willLPD2I

PR #222, opened 2022-12-08
file/folder checker

willLPD2I:
Function in the CinValidationSession class that checks if the input is a directory or a file, and what file type it is


cin_validator/cin_validator_class.py
```diff
@@ -10,18 +14,42 @@
...
+        fn = self.filename.name
+        if os.path.isfile(fn):
+            extension = pathlib.Path(fn).suffix
+            if extension[:4] == ".xml":
+                fulltree = ET.parse(fn)
+                root = fulltree.getroot()
+                self.data_files_obj = DataContainerWrapper(XMLtoCSV(root))
+                print("got here!")
+            elif extension[:4] == ".csv":
+                print(".csv support coming soon!")
```
tab1tha, line 36: If the file is a csv, I wonder if we are able to read the filename (or if it is safe to assume that the file names will be what we expect them to be).
A reliable means of processing csv files will probably be to read their column names and deduce the name of the file (table). For example, a file that contains CPPstartDate is probably the ChildProtectionPlans table. 
Coding this up might be too verbose but first let's make it work and then we make it better.

I understand that we have listed out the columns for each table in two places already (api.py and ingress.py). It won't be nice to do that a third time. Let's figure that out.

cin_validator/cin_validator_class.py
```diff
@@ -10,18 +14,42 @@
...
+            elif extension[:4] == ".zip":
+                zf = zipfile.ZipFile(fn)
+                dfs = {
+                    text_file.filename[:-4]: pd.read_csv(zf.open(text_file.filename))
+                    for text_file in zf.infolist()
+                    if text_file.filename.endswith(".csv")
```
tab1tha, line 42: I like that you checked you added the check for csv. I also wouldn't expect an xml file to be in a zip here.
Should we flag it if an xml is included among the csvs?
or simply pick out only the csvs?
or consider the xml only (since all tables can be generated from it)?

cin_validator/cin_validator_class.py
```diff
@@ -10,18 +14,42 @@
...
+                }
+                print(dfs)
+                self.data_files_obj = DataContainerWrapper(dfs)
+            else:
+                print(f"Filetype: {extension} not currently supported")
+        elif os.path.isdir(fn):
```
tab1tha, line 48: Getting a list of all the files in a directory such that you can check their types and read them (as you did above for the zip)
https://stackoverflow.com/questions/3207219/how-do-i-list-all-files-of-a-directory
https://www.geeksforgeeks.org/how-to-read-multiple-text-files-from-folder-in-python/

tab1tha (approved):
Hi Will, 

I think that you've thought through quite a number of the use cases. This looks good. 
A few things have changed in the cin_validator_class file including the fact that the data processing and validation have now been isolated. 

The changes here best fit into the data processing part now. So a few things here need to be moved around.
This is good work !

cin_validator/cin_validator_class.py
tab1tha: We can leave out directory handling since we won't be handling CSVs in the MVP version.

willLPD2I:
@tab1tha I realise I've forgotten, are we still going ahead with this?

tab1tha:
It can be closed down.  
If we need it later then the pull request can be reopened.

PR #233, opened 2022-12-13
8873

### LINKED ISSUE #71: Rule 8873 : When there is only one assessment on the episode and the factors code “21 No factors identified” has been used for the completed assessment, the reason for closure ‘RC8’ must be used.
Within a \<CINDetails\> group, if there is only one \<Assessment\> group present and \<AssessmentFactors\> (N00181) = “21”, \<ReasonForClosure\> (N00103) must = RC8.

willLPD2I:
closes #71

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8873.py
+++ b/cin_validator/rules/cin2022_23/rule_8873.py
@@ -4,7 +4,6 @@ import pandas as pd
 
 from cin_validator.rule_engine import CINTable, RuleContext, rule_definition
 from cin_validator.test_engine import run_rule
-from cin_validator.utils import make_census_period
 
 # Get tables and columns of interest from the CINTable object defined in rule_engine/__api.py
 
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8873.py
```diff
@@ -0,0 +1,290 @@
...
+AssessmentFactors = Assessments.AssessmentFactors
+CINdetailsID = Assessments.CINdetailsID
+AssessmentActualStartDate = Assessments.AssessmentActualStartDate
+
+CINdetails = CINTable.CINdetails
+LAchildID = CINdetails.LAchildID
+CINdetailsID_cin = CINdetails.CINdetailsID
+ReasonForClosure = CINdetails.ReasonForClosure
+
+# Reference date in header is needed to define the period of census.
```
tab1tha: Removed this comment since it is no longer needed.

cin_validator/rules/cin2022_23/rule_8873.py
```diff
@@ -0,0 +1,290 @@
...
+    df_ass_merged = df_ass_merged[
+        (
+            df_ass_merged["AssessmentActualStartDate_x"]
+            != df_ass_merged["AssessmentActualStartDate_y"]
+        )
+    ]
+    more_than_1_ass = df_ass_merged["ROW_ID_x"].tolist()
+
+    df_ass = df_ass[~df_ass["ROW_ID"].isin(more_than_1_ass)]
+
```
tab1tha, line 64: This works well. Another way you could have done it is to groupby the IDs and create a column of the count of Assessments per group.  Then filter based on the rows whose count is more than one. like this:

use transform() in the groupby to ensure that the series created in the result has the same dimension as other columns in the dataframe so that it fits in nicely. That is, the count value will be expanded such that the same count value is repeated for each row in a CIN group.
`df_ass["num_ass"] = df_ass.groupby(["LAchildID", "CINdetailsID"])["AssessmentActualStartDate"].transform("count")`

filter
`df_ass = df_ass[df_ass["num_ass"]<2]`

Now we have a df_ass with only the groups where 1 assessment occurs.

cin_validator/rules/cin2022_23/rule_8873.py
```diff
@@ -0,0 +1,290 @@
...
+                "ReasonForClosure": "RC8",
+                "CINdetailsID": "cinID1",
+            },
+            {  # 1 ignored
+                "LAchildID": "child1",
+                "ReasonForClosure": "RC8",
+                "CINdetailsID": "cinID2",
+            },
+            {  # 2 pass
+                "LAchildID": "child2",
```
tab1tha, line 183: Your test data is good. The comments I added were to make it more obvious so that it is easier to maintain. 
I also adjusted the "ignore" test cases such that they would actually have failed if they hadn't been ignored.

cin_validator/rules/cin2022_23/rule_8869.py
```diff
@@ -0,0 +1,101 @@
+from typing import Mapping
```
tab1tha, line 1: I notice that 8869 was mistakenly included here so I'll merge in main to prevent any conflicts before merging this in.

tab1tha (approved):
Hi Will. 
This is good work. It was interesting to see how you used a merge to count group items. I hadn't thought of doing it that way. 
I've included another way of doing it also.

I think this is good to go. So I'll do some housekeeping and merge it in.

PR #242, opened 2022-12-16
8815 working

### LINKED ISSUE #53: Rule 8815 : More than one open CIN Details episode (a module with no CIN Closure Date) has been provided for this child and case is not a referral with no further action.
If more than one \<CINdetails\> module is present for the same child, then there must only be one where:

\<CINclosureDate\> (N00102) is missing; and

\<ReferralNFA\> (N00112) = false or 0

Notes:  a) \<CINclosureDate\> (N00102) can be present on all modules, b) there may be more than one module with no \<CINclosureDate\> where \<ReferralNFA\> (N00112) = true/1.

willLPD2I:
closes #53


tab1tha (approved):
Nice. 
Good to go.

PR #243, opened 2022-12-19
8831

### LINKED ISSUE #56: Rule 8831 : Activity is recorded against a case marked as a referral with no further action
If a \<CINdetails\> module has \<ReferralNFA\> (N00112) = true or 1, then it cannot have any of the following:

\<AssessmentActualStartDate\> (N00159)
\<AssessmentAuthorisationDate\> (N00160)
\<S47ActualStartDate\> (N00148)
\<DateOfInitialCPC\> (N00110)

willLPD2I:
closes #56

### DIFF
--- a/Documentation/rule_logic_demos/rule8831_table_checks.ipynb
+++ b/Documentation/rule_logic_demos/rule8831_table_checks.ipynb
@@ -1,18 +1,16 @@
 {
  "cells": [
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "c98f9306",
+   "cell_type": "markdown",
+   "id": "df214fb6",
    "metadata": {},
-   "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


tab1tha (approved):
Hi Will, 
I've done more changes in this one. It has been more of reshuffling your code.
- The childProtectionsTable was included but it wasn't needed by the rule so it has been taken out. (The types of LAchildID and CINdetailsID can be defined from any table that has them so I used the CINdetails table instead.)
- Also, it is a bit risky to link three tables together because data can be lost when the two-table merges are merged to each other. We were able to avoid that here because the tables being compared could be compared to the cin table independent of each other. So I moved that around.
- I've added a Jupyter notebook in the commit that demonstrates my changes and how they fit into yours.

tab1tha:
I also shuffled expected_df around so that it fits with the result of the concat done on cin_issues. I haven't done a sort because we simply need to check that the values are produced and not the order which they're produced.

PR #250, opened 2022-12-20
8540 working

### LINKED ISSUE #30: Rule 8540 : Child’s disability is missing or invalid (see Disability table)
If \<PersonBirthDate\> (N00066) is present and at least one \<ReferralNFA\> (N00112) is false or 0, then one or more instances of \<Disability\> (N00099) must be present, and all instances must be valid code values

willLPD2I:
closes #30 

@tab1tha I had to finagle some of the df_issues merges here, the rule works via the pytest and I don't think it'll break anything but it might be worth a look.


tab1tha (approved):
This is well done. 

It responded to my test case changes as expected. Nice work Will !

PR #251, opened 2022-12-21
8816

### LINKED ISSUE #54: Rule 8816 : An open CIN episode is shown and case is not a referral with no further action, but it is not the latest episode.
If there is a \<CINdetails\> module present for the child where:

\<CINclosureDate\> (N00102) is missing; and

\<ReferralNFA\> (N00112) = false or 0

then the \<CINreferralDate\> (N00100) for this module must be the latest of all Referral Dates for that child.  

Note: This rule should be evaluated at Child-level for imported data.

willLPD2I:
closes #54

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8816.py
+++ b/cin_validator/rules/cin2022_23/rule_8816.py
@@ -4,7 +4,6 @@ import pandas as pd
 
 from cin_validator.rule_engine import CINTable, RuleContext, rule_definition
 from cin_validator.test_engine import run_rule
-from cin_validator.utils import make_census_period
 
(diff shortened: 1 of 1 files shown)


tab1tha (approved):
Hi Will, 

Here, the tweaks to the test data didn't respond as expected so I had to take the logic apart. 
I'll push the changes now.

tab1tha:
Using .transform in a `groupby` statement has enabled us to create a column where all the children in the group are assigned the `groups` value. an example is shown in the screenshot of what happens if we want to compare` DateOfInitialCPC` values to the maximum value recorded, per child.

when `idxmax` is used instead of `max` , the resulting column will contain the index positions of the maximum values instead.

image

PR #253, opened 2022-12-21
8866 working

### LINKED ISSUE #39: Rule 8866 : Source of Referral is missing or an invalid code
If \<CinReferralDate\> (N00100) is on or after 1 April 2013 then \<ReferralSource\> (N00152) must be present and must be a valid code

willLPD2I:
closes #39


tab1tha (approved):
Apart from comment updates, I reset the index (to save ROW_ID) before the filter because it is safer to do that before than after filtering.

PR #254, opened 2022-12-21
2884

### LINKED ISSUE #90: Rule 2884 : An initial child protection conference is recorded at both the S47 and CIN Details level and it should only be recorded in one
If present, the CIN Details module \<DateOfInitialCPC\> (N00110) must not be the same as any Section47 module \<DateOfInitialCPC\> (N00110).

willLPD2I:
closes #90


tab1tha (approved):
Works right out of the box.

PR #261, opened 2022-12-22
Rule8772

GBurrel:


willLPD2I:
@GBurrel I'll take a look at these later, I have a lot of meetings this morning, but I'll get back to you ASAP.

GBurrel:
No worries @willLPD2I - I appreciate it may have to wait till later.

willLPD2I:
Hi @GBurrel, I stared at this for a long time, and then stared at it so long I needed Tambe to stare at it with me too. I fixed a couple of little things and the error that was causing your current assertion error, you were doing your check on CINdetails so you only needed to return ReferralNFA as a failing column. I've left the rest to do which is sorting out the correct number of failing rows, and the correct expected df of failing rows, but if you feel short on time, I'm happy to wrap that up.

GBurrel:
Thanks @willLPD2I - I'll look at the failing rows and look to amend that to get it working. Once I've done that do I just commit the changes and let you know?

tab1tha:
yes @GBurrel.
That works, Thank you.

Also, if you run `black cin_validator` in your command line before committing your changes, it will fix the formatting so that the linting check passes.  If that doesn't work, don't worry about it. we can fix it from here.

tab1tha:
Thank you @GBurrel 

I've looked at this. It works now.

PR #264, opened 2022-12-22
2991Q working

### LINKED ISSUE #96: Rule 2991Q : Please check: A Section 47 module is recorded and there is no assessment on the episode
If \<Section47\> module is present then \<Assessment\> module should be present

willLPD2I:
closes #96 

@tab1tha I've done this in such a weird way I'm really sorry


tab1tha:
Hi @willLPD2I , you were right about this. There is, in fact, an easier way to do it.
You can put on the indicator in the merge and select the rows based on the value of the created `_merge` column.

PR #267, opened 2022-12-22
8898 working

### LINKED ISSUE #67: Rule 8898 : The assessment has more than one parental or child factors with the same code
If there is more than one \<AssessmentFactors\> (N00181) for an assessment recorded, then none of the values should appear more than once for an assessment.

willLPD2I:
closes #67


tab1tha:
Here are the changes https://github.com/data-to-insight/CIN-validator/blob/main/cin_validator/rules/cin2022_23/rule_8898.py
They were accidentally pushed to main.

PR #277, opened 2023-01-10
Process date columns

### LINKED ISSUE #271: Process date columns.
#### goal
Convert the datatype of all columns containing dates to datetime format.

#### affected files
ingress.py
utils.py

#### description
Create a function in utils.py that takes in a dataframe, selects columns that have the word "date" in them, converts those columns to datetime format, and returns the dataframe containing the converted columns.

In the ingress.py process, find where all the tables that are created, and just before they are saved, run the function on them to convert the appropriate columns to datetime format.

#### check
The `run-all `command should no longer contain errors of the form `TypeError, '>=' not supported between instances of 'str' and 'Timestamp'`

#### test
In the outer `tests` folder, create a `test_utils.py` file and write a test for the function you created earlier in utils. We can do this part together.

willLPD2I:
closes #271 

process_date_columns function added to utils, it's used every time a  child is made in ingress, but this might be slow, is it possible to do it outside of the create_table functions instead?

It clears up errors where we can't compare str and timestamp (yay), but I think it might make different errors now (boo!)

Still need to write a test!


willLPD2I:
I've combed through and found the new errors thrown by the function:

2886Q : TypeError, unsupported operand type(s) for &: 'DatetimeArray' and 'DatetimeArray'
8875: AttributeError, Can only use .dt accessor with datetimelike values

cin_validator/ingress.py
```diff
@@ -9,7 +9,7 @@
 import pandas as pd
 
 # TODO make this work with from cin_validator.utils import get_values
-from .utils import get_values, make_census_period
+from .utils import get_values, make_census_period, process_date_columns
```
tab1tha, line 12: Side note: the TODO above now works, you can replace ".utils" with "cin_validator.utils" and delete line 11.

cin_validator/rules/cin2022_23/rule_2886Q.py
```diff
@@ -30,8 +30,8 @@ def validate(
     num_records = len(df)
 
     # get the number of child records that fit the specified condition.
-    missing_gender = df[df[GenderCurrent].isna() | df[GenderCurrent] == 0]
-    missing_date = df[df[ExpectedPersonBirthDate].isna()]
+    missing_gender = df[GenderCurrent].isna() | df[GenderCurrent] == 0
+    missing_date = df[ExpectedPersonBirthDate].isna()
```
tab1tha, line 34: Good catch !
I wonder how its test passed.

cin_validator/utils.py
```diff
@@ -94,3 +94,13 @@ def __deepcopy__(self, memo):
...
+
+
+def process_date_columns(df):
+    for column in df:
+        if ("date" in column) | ("Date" in column):
+            try:
+                df[column] = df[column].apply(pd.to_datetime, format="%d/%m/%Y")
+            except:
+                df[column] = df[column].apply(pd.to_datetime, format="%Y/%m/%d")
+    return df
```
tab1tha, line 106: Sufficient !
This works. We now have to find a good place to run it from.

cin_validator/tinkering.py
```diff
@@ -0,0 +1,80 @@
+import pandas as pd
```
tab1tha, line 1: Was this your "trying_things" file?
Did you intentionally add it to the commit?

tab1tha (changes requested):
The logic works well.

Now, we need to find a more efficient way to run the function on all the tables.

tab1tha:
That's true. 
I didn't notice that putting it in per line will call the function every time a table is created for every child. It is more efficient to do it after all the tables have been created fully for all children present in the data.

Could you find a more central place to put it?

Suggestion one:
It could  be added to the init function of the class such that it runs after the "for every child" for loop.

Suggestion two:
How about we ensure that all XML to CSV conversions are managed in the same place and in that place, the function is run on all tables that have been generated by the XMLtoCSV class, before any further processing continues?.
The changes in the rpc-layer branch involve a process-data function in the cin_validator_class file. The data conversion is done there. Maybe we could do that in that function.

Suggestion three:
Where else could it be?...

cin_validator/tinkering.py
willLPD2I: oops haha, I tend to do my comit as ./cin_validator so it adds everything

tab1tha (approved):
Good to go.

PR #281, opened 2023-01-11
Refactor/integration fixes - Json Serialiser

tab1tha:
The changes in this pull request
- implement a json serialiser using [json.JSONEncoder](https://docs.python.org/3/library/json.html#json.JSONEncoder) that converts np.nan values to None values such that they can be typecasted into json null.
- adjust the version numbers of numpy and pandas to make them compatible with pyodide using `pip install --upgrade numpy==1.22.4` and `pip install --upgrade pandas==1.4.2` respectively and then running `poetry lock` and `poetry install`.
- separate parsing of files received from the frontend(RemoteFile object) and those received when using the command line interface (TextIOwrapper object).

### DIFF
--- a/cin_validator/__main__.py
+++ b/cin_validator/__main__.py
@@ -6,7 +6,6 @@ import click
 import pytest
 
 from cin_validator import cin_validator_class as cin_class
-from cin_validator.ingress import XMLtoCSV
 from cin_validator.rule_engine import registry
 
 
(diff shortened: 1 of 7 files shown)


tab1tha:
The main branch was merged in to resolve a possible merge conflict.

dotloadmovie (approved):

willLPD2I (approved):

PR #289, opened 2023-01-18
feature/select rules

tab1tha:
The changes in this pull request allow the a sub-list of the existing rules to be run on the data, based on the user's preference.
A `selected_rules` argument has been added to the `cin_validate` api function. It can accept a single string value or a list of strings where each string represents a rule code.

Other changes include further modularising of code to split up the large for loop that existed in the `create_error_report` function.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -40,7 +40,9 @@ The Header contains the metadata including collection_year and census type. If C
 ## Validation: cin_validate
 This is the default feature. File input is converted into tabular-format (dataframes) and the data is validated based on rules [defined by the DfE](https://www.gov.uk/government/publications/children-in-need-census-2022-to-2023-specification) and coded in the `rules` folder.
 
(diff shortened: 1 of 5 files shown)


willLPD2I:
@tab1tha At first glance this looks good, do you mind if I go and change some of the formatting of docstrings for consistency with the ones I wrote? Just things like having capitals at the start of sentences and having line breaks between function descriptions and parameters for readability?

willLPD2I (commented):
Yep, looks good, but as per my comment I'd like to format some of the docstirngs a little. I'm happy to do it?

dotloadmovie (approved):

tab1tha:
> Yep, looks good, but as per my comment I'd like to format some of the docstirngs a little. I'm happy to do it?

Yes. Sure @willLPD2I 

If you can get it in by the end of today, that works.
Else, I can merge this in so that the frontend part can start being built and then you open a new pull request when your changes are ready. I'll wait until close of business today.

tab1tha:
Merging this in as agreed.

PR #293, opened 2023-01-19
Fix dateofinitialcpc bug

### LINKED ISSUE #292: Key error: DateOfInitialCPC error in run-all
when running all against fake date, there is a key error for some tables with  DateOfInitialCPC

willLPD2I:
closes #292 

Tables both containing DateOfInitialCPC were merged, not on DateOfInitialCPC, but then DateOfInitialCPC was used without suffixes. DateOfInitialCPC was also not present in sample dataframes allowing tests to pass.

DateOfInitialCPC added to sample DFs and Merges correctly done using DateOfInitialCPC.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_2990.py
+++ b/cin_validator/rules/cin2022_23/rule_2990.py
@@ -69,7 +69,7 @@ def validate(
     )
 
     df_cin_47 = df_cin.merge(
-        df_47, on=["LAchildID", "CINdetailsID"], how="left", suffixes=["_cin", "_47"]
+        df_47, on=["LAchildID", "CINdetailsID", DateOfInitialCPC], how="left", suffixes=["_cin", "_47"]
     )
 
     df_cin_cin_pd = df_cin.merge(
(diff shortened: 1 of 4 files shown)


tab1tha (changes requested):
Nice catch Will. This is good work. 
There's a bit of a caveat though. The DateOfInitialCPC in the CINdetails table shouldn't be equal to that in the Section47 table https://github.com/data-to-insight/CIN-validator/issues/90 so if you join on DateOfInitialCPC in real world data, no matches will be found. 

Another method could be to allow suffixes to be added to DateOfInitialCPC and use it with the suffixes for the rules where it is required. Then it can be converted from the suffix form to its real name before pushing the error locations per table involved.

willLPD2I:
> Nice catch Will. This is good work. There's a bit of a caveat though. The DateOfInitialCPC in the CINdetails table shouldn't be equal to that in the Section47 table #90 so if you join on DateOfInitialCPC in real world data, no matches will be found.
> 
> Another method could be to allow suffixes to be added to DateOfInitialCPC and use it with the suffixes for the rules where it is required. Then it can be converted from the suffix form to its real name before pushing the error locations per table involved.

@tab1tha Well it's good you knew this because I did not! I'll get to it now. Finding the issue was more of a problem than the fix anyway

tab1tha (approved):

tab1tha:
This was a tricky bug. Thank you, Will. 
Your fix works.

PR #295, opened 2023-01-20
Fix 8816 - ValueError: attempt to get argmax of an empty sequence

tab1tha:
Children with no CINreferralDate raised this error. The fix filters out nan values before finding the latest referral such that the maximum value of an empty sequence never needs to be found.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8816.py
+++ b/cin_validator/rules/cin2022_23/rule_8816.py
@@ -142,6 +142,20 @@ def test_validate():
                 "CINdetailsID": "cinID3",
                 "ReferralNFA": "0",
             },
+            {
+                "LAchildID": "child4",
+                "CINclosureDate": pd.NA,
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8816.py
```diff
@@ -45,6 +45,9 @@ def validate(
     # <ReferralNFA> (N00112) = false or 0
     # then the <CINreferralDate> (N00100) for this module must be the latest of all Referral Dates for that child.
 
+    # removing nans prevents ValueError: attempt to get argmax of an empty sequence, when no CINreferralDate is present
+    df_cin = df_cin[df_cin[CINreferralDate].notna()]
```
willLPD2I, line 49: Good idea, I didn't think of this because I was worried about this meaning there could be multiple empty CInreferralDates, but re-reading the logic, that doesn't matter! Good fix!

PR #296, opened 2023-01-20
diagram.data flow in cin tool.

tab1tha:
powerpoint was converted to a pdf document so that it is displayed by default. The powerpoint can be shared on request through slack.


tab1tha:
This isn't visible until it's merged. 
Still, suggest any changes if any.

PR #300, opened 2023-01-26
Feature/include bank holidays

tab1tha:
The changes update rules 8569, 8670Q and 8675Q such that UK bank holidays are always considered in their calculations of number of working days between dates.
It uses https://www.gov.uk/bank-holidays and numpy to do this.

### DIFF
--- a/cin_validator/utils.py
+++ b/cin_validator/utils.py
@@ -1,6 +1,8 @@
 from copy import deepcopy
 
+import numpy as np
 import pandas as pd
+from govuk_bank_holidays.bank_holidays import BankHolidays
 
 
 def get_values(xml_elements, table_dict, xml_block):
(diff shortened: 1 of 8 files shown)


cin_validator/utils.py
```diff
@@ -1,6 +1,8 @@
 from copy import deepcopy
 
+import numpy as np
 import pandas as pd
+from govuk_bank_holidays.bank_holidays import BankHolidays
```
willLPD2I, line 5: wow this is really cool! Who maintains this? I'm so glad it exists!

willLPD2I (approved):

cin_validator/utils.py
tab1tha: Same here. 
I was quite happy to see it. It is maintained by the ministry of Justice https://github.com/ministryofjustice/govuk-bank-holidays

This is the pip page https://pypi.org/project/govuk-bank-holidays/

PR #304, opened 2023-01-31
Fix/no error scenario

tab1tha:
generate an empty report when no errors are found.


tab1tha:
Merging this in because I need these changes.

PR #307, opened 2023-02-03
added simple timer to CLI for testing purposes

willLPD2I:
adds timer that automatically runs suing fake_cin_data for code testing purposes

### DIFF
--- a/cin_validator/__main__.py
+++ b/cin_validator/__main__.py
@@ -5,6 +5,9 @@ from pathlib import Path
 import click
 import pytest
 
+import datetime
+import os
+
 from cin_validator import cin_validator_class as cin_class
 from cin_validator.rule_engine import registry
 
@@ -178,6 +181,14 @@ def cli_converter(filename: str):
             v.to_csv(filepath)
     else:
         click.echo(f"{filename} can't be found, have you entered it correctly?")
+
+@cli.command(name='timer')
+def timer():    
+    st = datetime.datetime.now()
+    os.system("python -m cin_validator run-all /workspaces/CIN-validator/fake_data/fake_CIN_data.xml")
+    et = datetime.datetime.now()
+    time_elapsed = et- st
+    print(f"Time to run: {time_elapsed}")
 
 
 if __name__ == "__main__":


tab1tha:
Hi @willLPD2I 
If you pull in changes from the main branch now, the lint check will be fixed.

cin_validator/__main__.py
```diff
@@ -183,5 +186,16 @@ def cli_converter(filename: str):
...
+@cli.command(name="timer")
+def timer():
+    st = datetime.datetime.now()
+    os.system(
+        "python -m cin_validator run-all /workspaces/CIN-validator/fake_data/fake_CIN_data.xml"
+    )
+    et = datetime.datetime.now()
+    time_elapsed = et - st
+    print(f"Time to run: {time_elapsed}")
+
```
tab1tha, line 198: This is useful.

Did you already consider allowing the filepath to be passed in as a command line argument?.

tab1tha (changes requested):

cin_validator/__main__.py
willLPD2I: Hi @tab1tha I did, I just realised we'd likely be using the same sample data every time, so there was no real reason to allow the filepath to be added as a CLI argument. I could potentially add it as an optional argument? I was going to do that originally actually but I was having a hard time getting one CLI function to work within this one.

tab1tha: Your filepath is a codespaces path and mine isn't so we'll still have to swap it out each time we run.

How about passing it in as a string and then appending that string to the `"python -m cin_validator run-all` string?
You could also use  a formatted-string format to put them together.

In any case, if this will take more than a moment to resolve, add a TODO statement above it and let it be. We can get to it later. There's a stack of tasks to prioritise.

willLPD2I: let's add it as a todo and come back later

willLPD2I:
@tab1tha I added the filepath option it took like 5 minutes, I forgot I could just use an f string

cin_validator/__main__.py
```diff
@@ -183,5 +186,15 @@ def cli_converter(filename: str):
         click.echo(f"{filename} can't be found, have you entered it correctly?")
 
 
+@cli.command(name="timer")
+@click.argument("filepath", type=str, required=True)
+def timer(filepath):
+    st = datetime.datetime.now()
+    os.system(f"python -m cin_validator run-all {filepath}")
```
tab1tha, line 193: Exactly !
A formatted string does the job. 

This is good to go.

tab1tha (approved):

PR #325, opened 2023-02-07
1560Q now correctly checks <FormerUPN> not <UPN>

### LINKED ISSUE #309: Bug: Rule 1560Q - FLags UPN as wrongly formatted when it should check former upn
**Expected behavior**
Should flag Former UPN

**Screenshots**
NOTE: Ensure you remove any identifying information

**Proposed fix (Optional)**
Check former upn instead of upn

willLPD2I:
closes #309


tab1tha (approved):
Nice catch.
This is good to go.

PR #326, opened 2023-02-07
4180 bugfix

### LINKED ISSUE #312: Bug: Rule 4180 - Marks genders as missing when gender is present
**Describe the bug**
Flags as gender missing when gender has a numerical value

willLPD2I:
4180 now checks gender as int or string. I'd rather sanitise the inputs converting to int but I didn't want to mess with ingress, and as I can't test as run-all isn't working, I did this. It might not even work but as I can't test with my data, it's a start.

closes #312

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4180.py
+++ b/cin_validator/rules/cin2022_23/rule_4180.py
@@ -40,7 +40,8 @@ def validate(
     # <GenderCurrent> (N00097) must be present and valid
 
     failing_indices = df[
-        df[GenderCurrent].isna() | (~df[GenderCurrent].isin(valid_gender_codes))
+        df[GenderCurrent].isna()
+        | (~df[GenderCurrent].astype("str").isin(valid_gender_codes))
     ].index
 
     # Replace ChildIdentifiers and GenderCurrent with the table and column name concerned in your rule, respectively.
--- a/cin_validator/rules/cin2022_23/rule_4180.py
+++ b/cin_validator/rules/cin2022_23/rule_4180.py
@@ -35,9 +35,10 @@ def validate(
 
     # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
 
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_4180.py
```diff
@@ -35,9 +35,10 @@ def validate(
 
     # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
 
-    valid_gender_codes = [1, 2, 0, 9]
+    valid_gender_codes = [1, 2, 0, 9, "1", "2", "0", "9"]
 
     # <GenderCurrent> (N00097) must be present and valid
+
     failing_indices = df[
         df[GenderCurrent].isna() | (~df[GenderCurrent].isin(valid_gender_codes))
```
tab1tha, line 43: you could also do 
`~df[GenderCurrent].astype("str").isin(valid_gender_codes)` and delete the integer values of valid_gender_codes.

tab1tha: In that way, whatever format the input has, you can be sure that it'll be a string when compared to your string expected values.

use `.astype(str)` - no quotes - if the above line raises an error.

willLPD2I: ah, I was doing astype int which was causing an error, I'll try int now

cin_validator/rules/cin2022_23/rule_4180.py
```diff
@@ -35,11 +35,13 @@ def validate(
 
     # implement rule logic as described by the Github issue. Put the description as a comment above the implementation as shown.
 
-    valid_gender_codes = [1, 2, 0, 9]
+    valid_gender_codes = [1, 2, 0, 9, "1", "2", "0", "9"]
```
tab1tha: Since you are doing df.astype(str), you can expect all values at the comparison stage to be in string format.

tab1tha (approved):

PR #330, opened 2023-02-09
bugfix 8869

### LINKED ISSUE #321: Bug: Rule 8869 - flags assessment factor as using 21 with others when the value is $A (unless there's another module not showing)
**Describe the bug**
Should flag where assessment factor 21 is used in conjunction with others (I'm going to look at the data again to check this one, my notes don't make sense.)

willLPD2I:
closes #321 
8869 wrongly flagged rows where children had 21 as an assessment factor in a different assessment. This was because it checked assessment factors within one CINplan, where CIN plans can have multiple assessment periods. This was fixed by using AssessmentAuthorisationDate in the merge to separate out different assessment modules. It would be possible to merge on CINdetailsID and AssessmentAuthorisationDate but it's redundant.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8869.py
+++ b/cin_validator/rules/cin2022_23/rule_8869.py
@@ -85,7 +85,7 @@ def test_validate():
         "AAAA",  # id4, cinid4 fail. same assessment period has factor 21
         "AA",
         "21",
-        "14",
+        "14",  # ignored. not the same auth date as preceding.
         "15",
     ]
     # non-date placeholders can be used since this rule doesn't require date-object properties.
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8869.py
```diff
@@ -50,11 +50,14 @@ def validate(
     # If <AssessmentFactors> (N00181) = “21”, this must be the only <AssessmentFactors> (N00181) present.
     df_orig = df.copy()
 
-    df = df[df[AssessmentFactors] == "21"]
+    df = df[(df[AssessmentFactors] == "21") | (df[AssessmentFactors] == 21)]
```
tab1tha: We can use astype here.

cin_validator/rules/cin2022_23/rule_8869.py
```diff
@@ -50,11 +50,14 @@ def validate(
...
 
     #  Merge tables
     df = df.merge(
-        df_orig, how="left", on=["LAchildID", "CINdetailsID"], suffixes=["", "_orig"]
+        df_orig,
```
tab1tha, line 58: This assumes that no two cin episodes of  a child can ever have the same authorization date. Which should be true in real life but might happen in the data.
I say it doesn't hurt to leave the check in.

tab1tha (approved):

PR #331, opened 2023-02-09
Test/general tests

tab1tha:
Check that the ruleset has the number of rules expected.

### DIFF
--- a/tests/test_ruleset.py
+++ b/tests/test_ruleset.py
@@ -1,0 +1,5 @@
+def test_ruleset_complete():
+    import cin_validator.rules.cin2022_23
+    from cin_validator.rule_engine import registry
+
+    assert len(registry) == 107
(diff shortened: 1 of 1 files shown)


willLPD2I (changes requested):
Can you get this documented in sphinx format like the rest of the code then it looks good to go!

tab1tha:
> Can you get this documented in sphinx format like the rest of the code then it looks good to go!

Sure

tab1tha:
> > Can you get this documented in sphinx format like the rest of the code then it looks good to go!
> 
> Sure

Actually, it is a test function. So it neither has parameters nor does it return anything. @willLPD2I 
With the test functions, can we include explanatory documentation only if needed ?

willLPD2I:
> > > Can you get this documented in sphinx format like the rest of the code then it looks good to go!
> > 
> > 
> > Sure
> 
> Actually, it is a test function. So it neither has parameters nor does it return anything. @willLPD2I With the test functions, can we include explanatory documentation only if needed ?

Sounds fair! If you could pop something really brief in the docs then?

tab1tha:
When changes from the main branch are brought in, this test now fails because rules 8899Q and 8617 have been removed (https://github.com/data-to-insight/CIN-validator/issues/344).
Hence the new total number of tests should be 105 instead of 107.

PR #335, opened 2023-02-10
removes select-issue-by-id feature

tab1tha:


tab1tha:
The changes in this pull request remove the feature that allows the user to filter the issue report by error ID. This is suggested for the following reasons:
- It is not a key functionality. It was created to simulate the concept of highlighting related issue locations at once in the frontend. However, it cannot be built upon because that will involve a frontend-backend call every time the user clicks on a data point in her view. That's not efficient.
- The assumption is that all elements of the ERROR_ID are strings. However, this has been found to not be true. Elements are in the datatype that was appropriate for computation in their respective rules. That means, sometimes digits in the ERROR_ID could be integers and other time they could be strings or floats and still look the same. This explodes the number of checks that would be necessary to parse the issue_id input into a format that can be found in the dataframe of issues.

In summary, this feature cannot be built on, is only applicable to a narrow use case as-is (when all the elements in the ERROR_ID are strings), and runs the risk of introducing technical debt, if left to exist.

willLPD2I (approved):

PR #336, opened 2023-02-13
bugfix 8890

### LINKED ISSUE #322: Bug: Rule 8890 -  flags concurrent s47s when module only has one s47ActualStartDate

willLPD2I:
closes #322 

Given that it should flag:
 b) the <S47ActualStartDate> (N00148) and the <ReferenceDate> (N00603) of any other <Section47> group  that has a missing <DateOfInitialCPC> (N00110) and the <ICPCnotRequired> (N00111) flag is not true,
this rule should not flag as concurrent s47 modules where ICPCnotRequired is true, which it is, which is causing rows to be wrongly flagged, eg:

```
                <Section47>
                    <S47ActualStartDate>xxx</S47ActualStartDate>
                    <ICPCnotRequired>true</ICPCnotRequired>
                </Section47>
                <Section47>
                    <S47ActualStartDate>xxx</S47ActualStartDate>
                    <ICPCnotRequired>true</ICPCnotRequired>
                </Section47>
```

The changes mean that modules where ICPCnotrequired == true are not picked up.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8890.py
+++ b/cin_validator/rules/cin2022_23/rule_8890.py
@@ -67,10 +67,10 @@ def validate(
 
     #  Create dataframes which only have rows with s47 modules, and which don't have ICPCnotrequired as true should have one plan per row.
     df_47 = df_47[
-        (df_47[S47ActualStartDate].notna()) & (df_47[ICPCnotRequired] != "true")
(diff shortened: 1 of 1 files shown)


tab1tha (approved):
The bug was caused because the rule said we should check for "true" but the code checked "1".
That has been resolved now.

tab1tha (approved):

PR #337, opened 2023-02-13
bugfix 4001, added check for CINplanStartDate

### LINKED ISSUE #311: Bug: Rule 4001 - Flags concurrent CIN and CPP plan but FE does not show CIN plan dates module
**Describe the bug**
Flags concurrent CIN and CPP plan but FE does not show CIN plan dates module so you can't check if it should be doing this. Given the ESCC data is meant to be perfect, it porbbaly shouldn't be highlighting it at all

willLPD2I:
closes #311 

If a <CINDetails> module has a <ChildProtectionPlan> module present with no <CPPendDate> (N00115)
- then a <CINPlanDates> module with no <CINPlanEndDate> (N00690) must not be present
The current implimentation of this rule is flagging children as failing who do not have a CINplanDates module, so they should not be failing., eg:

```
<CINdetails>
                <CINreferralDate>XXXX</CINreferralDate>
                <ReferralSource>XX</ReferralSource>
                <PrimaryNeedCode>XX</PrimaryNeedCode>
                <ReferralNFA>X</ReferralNFA>
                <ChildProtectionPlans>
                    <CPPstartDate>XXXX</CPPstartDate>
                    <InitialCategoryOfAbuse>XXX</InitialCategoryOfAbuse>
                    <LatestCategoryOfAbuse>XXX</LatestCategoryOfAbuse>
                    <NumberOfPreviousCPP>X</NumberOfPreviousCPP>
                    <Reviews>
                        <CPPreviewDate>XXX</CPPreviewDate>
                        <CPPreviewDate>XXX</CPPreviewDate>
                        <CPPreviewDate>XXX</CPPreviewDate>
                    </Reviews>
                </ChildProtectionPlans>
            </CINdetails>

```
Where this child should not fail as they have no CINplanDates module

The fix adds a check for CINplanStartDate so that it won't fail rows where there is no CINplan at all

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4001.py
+++ b/cin_validator/rules/cin2022_23/rule_4001.py
@@ -47,7 +47,9 @@ def validate(
     )
 
     #  Get rows where CPPendDate is null and CINPlanEndDate is null
-    condition = df_merged[CPPendDate].isna() & df_merged[CINPlanEndDate].isna()
+    condition = df_merged[CPPendDate].isna() & (
+        df_merged["CINplanStartDate"].notna() & df_merged[CINPlanEndDate].isna()
+    )
     df_merged = df_merged[condition].reset_index()
 
     df_merged["ERROR_ID"] = tuple(zip(df_merged[LAchildID], df_merged[CINdetailsID]))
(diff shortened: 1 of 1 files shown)


tab1tha (approved):
Good to go.

PR #338, opened 2023-02-14
Bugfix 8863

### LINKED ISSUE #319: Bug: Rule 8863 - Assessment flagged as ongoing when another is ongoing when FE only shows one in module
**Describe the bug**
Assessment flagged as ongoing when another is ongoing when FE only shows one in module, meaning that either the FE isn't showing the one that's in conflict, or there is only one assessment and it shouldn't be flagged.

willLPD2I:
closes #319 

Within one <CINdetails> group, where present each <AssessmentActualStartDate> (N00159) must not fall on or between:
    a) the <AssessmentActualStartDate> (N00159) and <AssessmentAuthorisation Date (N00160) of any other <Assessments> 
        group that has an <AssessmentAuthorisationDate> (N00160)
    OR
    b) the <AssessmentActualStartDate> (N00159) and the <ReferenceDate> (N00603) where the 
        <AssessmentAuthorisationDate> (N00160) is missing

This is incorrectly failing the following data:

```
                <Assessments>
                    <AssessmentActualStartDate>2021-09-XX</AssessmentActualStartDate>
                    <AssessmentAuthorisationDate>2021-10-XX</AssessmentAuthorisationDate>
                </Assessments>
                <Assessments>
                    <AssessmentActualStartDate>2021-03-XX</AssessmentActualStartDate>
                    <AssessmentAuthorisationDate>2021-04-XX</AssessmentAuthorisationDate>
                </Assessments>
```

Where both have assessment authorisation dates, and do not occur concurrently.

I was able to recreate the bug in the cod eby using the following in the sample df for 8863:
```

            {
                "LAchildID": "child5",
                "CINdetailsID": "cinID1",
                "AssessmentActualStartDate": "01/03/2000",  
                "AssessmentAuthorisationDate": "01/04/2000",
            },  
            {
                "LAchildID": "child5",
                "CINdetailsID": "cinID1",
                "AssessmentActualStartDate": "01/09/2000", 
                "AssessmentAuthorisationDate": "01/10/2000",
            },
```

On if we call the first Ass. period 1, and the second 2, and s for start, e for end, and r for reference date, the code was originally checking if 1s was after  2s and (correct) and if 1s was before 1e or the reference date (wrong). I've fixed it so it now checks if 1s is between 2s and 2e/r, which is right. I've illustrated the change with an image which may or may not make sense: 
image

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8863.py
+++ b/cin_validator/rules/cin2022_23/rule_8863.py
@@ -102,7 +102,7 @@ def validate(
     df_merged = df_merged[
         ass_started_after_start & (ass_started_before_end | ass_started_before_refdate)
     ].reset_index()
-    print(df_merged['LAchildID'])
+    print(df_merged["LAchildID"])
     # create an identifier for each error instance.
     df_merged["ERROR_ID"] = tuple(
         zip(
@@ -199,13 +199,13 @@ def test_validate():
             {
                 "LAchildID": "child5",
                 "CINdetailsID": "cinID1",
-                "AssessmentActualStartDate": "01/03/2000",  
+                "AssessmentActualStartDate": "01/03/2000",
                 "AssessmentAuthorisationDate": "01/04/2000",
-            },  
+            },
             {
                 "LAchildID": "child5",
                 "CINdetailsID": "cinID1",
-                "AssessmentActualStartDate": "01/09/2000", 
+                "AssessmentActualStartDate": "01/09/2000",
                 "AssessmentAuthorisationDate": "01/10/2000",
             },
         ]
--- a/cin_validator/rules/cin2022_23/rule_8863.py
+++ b/cin_validator/rules/cin2022_23/rule_8863.py
@@ -102,7 +102,7 @@ def validate(
     df_merged = df_merged[
         ass_started_after_start & (ass_started_before_end | ass_started_before_refdate)
     ].reset_index()
-    
+
     # create an identifier for each error instance.
     df_merged["ERROR_ID"] = tuple(
         zip(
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8863.py
```diff
@@ -88,21 +88,21 @@ def validate(
...
-        df_merged["AssessmentActualStartDate_ass"]
-        <= df_merged["AssessmentAuthorisationDate_ass"]
-    ) & df_merged["AssessmentAuthorisationDate_ass"].notna()
+        df_merged["AssessmentActualStartDate_ass"]  # 1 starts earlier than 2 finishes
+        <= df_merged["AssessmentAuthorisationDate_ass2"]
+    ) & df_merged["AssessmentAuthorisationDate_ass2"].notna()
     ass_started_before_refdate = (
         df_merged["AssessmentActualStartDate_ass"] <= reference_date
-    ) & df_merged["AssessmentAuthorisationDate_ass"].isna()
+    ) & df_merged["AssessmentAuthorisationDate_ass2"].isna()
```
tab1tha, line 100: This bug seems like it could've taken a while to find.
Good spot.

cin_validator/rules/cin2022_23/rule_8863.py
```diff
@@ -88,21 +88,21 @@ def validate(
...
 
     df_merged = df_merged[
         ass_started_after_start & (ass_started_before_end | ass_started_before_refdate)
     ].reset_index()
-
+    print(df_merged["LAchildID"])
```
tab1tha, line 105: to delete

tab1tha (changes requested):
remove print statement and merge.

tab1tha (approved):

PR #340, opened 2023-02-15
Bugfix 8820

### LINKED ISSUE #316: Bug: Rule 8820 - Flags CIN episode dates as overlapping when there's only one module
**Describe the bug**
Flags cin episode dates as overlapping when there's only one module (7038354) (flags CINref and CInclosure in one module), could be correct but the FE only shows one module? As the ESCC data is meant to be right, this is unlikely, however.

willLPD2I:
closes #316 

```
f more than one <CINdetails> module is present for the same child, then for each module the
 <CINreferralDate> (N00100) must not fall;
a) between the <CINreferralDate> (N00100) and the <CINclosureDate> (N00102) if present; and
 b) between the <CINreferralDate> (N00100) and the <ReferenceDate> (N00603)
 if the <CINclosureDate> is not present and <ReferralNFA> (N00112) = false or 0
```

The rule currently fails:
```
            <CINdetails>
                <CINreferralDate>2021-07-AA</CINreferralDate>
                <CINclosureDate>2021-08-XX</CINclosureDate>
                <ReferralNFA>false</ReferralNFA>
            </CINdetails>
            <CINdetails>
                <CINreferralDate>2021-05-XX</CINreferralDate>
                <CINclosureDate>2021-07-AA</CINclosureDate>
                <ReferralNFA>false</ReferralNFA>
            </CINdetails>
```

Where  the referral date of the first block is the same as the closure date of the second. To fix the rule, allow the referral and closure of blocks to be the same.

The fix removes the = signs from the >=/<= equality checks as these meant you couldnt finish a plan one day and starta  new one the same day, which is allowed.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8820.py
+++ b/cin_validator/rules/cin2022_23/rule_8820.py
@@ -87,23 +87,22 @@ def validate(
 
     # Determine overlaps
     cin_started_after_start = (
-        df_merged["CINreferralDate_cin"] >= df_merged["CINreferralDate_cin2"]
+        df_merged["CINreferralDate_cin"] > df_merged["CINreferralDate_cin2"]
     )
     cin_started_before_end = (
-        df_merged["CINreferralDate_cin"] <= df_merged["CINclosureDate_cin2"]
+        df_merged["CINreferralDate_cin"] < df_merged["CINclosureDate_cin2"]
     ) & df_merged["CINclosureDate_cin2"].notna()
 
-    falsezero = ["False", "0"]
     cin_started_before_refdate = (
-        (df_merged["CINreferralDate_cin"] <= reference_date)
+        (df_merged["CINreferralDate_cin"] < reference_date)
         & df_merged["CINclosureDate_cin2"].isna()
-        & df_merged["ReferralNFA_cin2"].isin(falsezero)
+        & df_merged["ReferralNFA_cin2"].isin(["false", "0", 0, "False"])
     )
 
     df_merged = df_merged[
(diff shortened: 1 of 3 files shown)


cin_validator/rules/cin2022_23/rule_8820.py
```diff
@@ -87,23 +87,22 @@ def validate(
...
+        df_merged["CINreferralDate_cin"] < df_merged["CINclosureDate_cin2"]
     ) & df_merged["CINclosureDate_cin2"].notna()
 
-    falsezero = ["False", "0"]
     cin_started_before_refdate = (
-        (df_merged["CINreferralDate_cin"] <= reference_date)
+        (df_merged["CINreferralDate_cin"] < reference_date)
         & df_merged["CINclosureDate_cin2"].isna()
-        & df_merged["ReferralNFA_cin2"].isin(falsezero)
+        & df_merged["ReferralNFA_cin2"].isin(["false", "0", 0, "False"])
```
tab1tha, line 99: all non-date columns that reach the rules can be trusted to be in string format. So we don't need to check for integers. 
Using `str.lower() ` also means we can be happy with all possible capitalisations of "False". 
`column_name.str.lower().isin(...)`

tab1tha (changes requested):
Only the 8820 rule has been reviewed. The rest will be reviewed in their respective pull requests.
Since these changes are minor, I can push them at the end of reviewing the trio. 

Still, if you get here before I do, feel free to do so.

tab1tha (approved):

PR #341, opened 2023-02-15
Bugfix 2885

### LINKED ISSUE #310: Bug: Rule 2885 - flags as CPP not on ICPC date when ICPCnotRequired is true
**Expected behavior**
If ICPCnot required is true, shouldn't flag the CPP not being on that date, as there is not one.

**Proposed fix (Optional)**
Don't fail rows with ICPCnotRequired as true

willLPD2I:
closes #310 

The following is being wrongly failed

  ```
              <Section47>
                    <DateOfInitialCPC>SAME DATE</DateOfInitialCPC>
                </Section47>
              <Section47>
              </Section47>
                <ChildProtectionPlans>
                    <CPPstartDate>SAME DATE</CPPstartDate>
                </ChildProtectionPlans>

```
It fails because there is an s47 module with no date, where if there is a module with a passing date, this child should pass and the no date s47 module should not matter.

The fix needed to fail CINdetails modules with children with no dates, and no correct dates, but allow modules with multiple s47s, some of which pass, to pass. This involves finding out is a CINdetails module has both s47 modules with, and without dates, and if that was the case, only testing the ones WITH dates. If these fail, the child still fails.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_2885.py
+++ b/cin_validator/rules/cin2022_23/rule_2885.py
@@ -99,8 +99,18 @@ def validate(
         suffixes=["_47", "_cin"],
         # the suffixes apply to all the columns not "merged on". That is, DateOfInitialCPC
     )
-    print(merged_df[['LAchildID', "DateOfInitialCPC_47", "DateOfInitialCPC_cin", "CPPstartDate", "CINdetailsID"]])
-    
+    print(
+        merged_df[
+            [
+                "LAchildID",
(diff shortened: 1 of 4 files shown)


willLPD2I:
okay oops, I switched branches using git siwtch -c branch name for this pr which messed up a little, as I forgot it takes work already done on the current branch...

tab1tha (approved):
Let's take this in as a hotfix for now. 
This might be a rule that should be rebuilt from scratch.

PR #352, opened 2023-02-22
Bugfix/fix merges

willLPD2I:
Fixed issues in 4000 and 8867 where they picked up and failed either children who didn't have a module and failed them as if they did. It was because the merges were either the wrong way round, or left instead of inner.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4000.py
+++ b/cin_validator/rules/cin2022_23/rule_4000.py
@@ -40,15 +40,16 @@ def validate(
         | (df_CINdetails[ReferralNFA] == True)
     ]
 
-    df_merged = df_cpd.merge(
-        df_CINdetails,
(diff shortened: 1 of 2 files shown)


cin_validator/rules/cin2022_23/rule_4000.py
```diff
@@ -40,12 +40,12 @@ def validate(
...
 
-    df_merged = df_cpd.merge(
-        df_CINdetails,
+    df_merged = df_CINdetails.merge(
+        df_cpd,
         left_on=["LAchildID", "CINdetailsID"],
         right_on=["LAchildID", "CINdetailsID"],
-        how="left",
-        suffixes=("_cpd", "_cin"),
+        how="inner",
```
tab1tha, line 47: Totally.

tab1tha (approved):

PR #357, opened 2023-02-22
fixed 8535Q

willLPD2I:
Allowed children who died to not have a birth date

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8535Q.py
+++ b/cin_validator/rules/cin2022_23/rule_8535Q.py
@@ -34,7 +34,7 @@ def validate(
 
     # df with all rows meeting the conditions
     df_issues = df[condition1].reset_index()
-    print(df_issues)
+
     link_id = tuple(
         zip(
             df_issues[LAchildID], df_issues[PersonDeathDate], df_issues[PersonBirthDate]
--- a/cin_validator/rules/cin2022_23/rule_8535Q.py
+++ b/cin_validator/rules/cin2022_23/rule_8535Q.py
@@ -31,12 +31,10 @@ def validate(
 
     # Return rows where DOB is prior to DOD
     condition1 = df[PersonBirthDate] > df[PersonDeathDate]
-    # Return rows with no DOB
-    condition2 = df[PersonBirthDate].isna()
 
     # df with all rows meeting the conditions
-    df_issues = df[condition1 | condition2].reset_index()
-
+    df_issues = df[condition1].reset_index()
+    print(df_issues)
     link_id = tuple(
         zip(
             df_issues[LAchildID], df_issues[PersonDeathDate], df_issues[PersonBirthDate]
@@ -57,7 +55,6 @@ def validate(
 
 
 def test_validate():
-
     child_identifiers = pd.DataFrame(
         [
             {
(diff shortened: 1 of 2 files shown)


tab1tha:
Oh apologies, I have read it again.
The first part of the rule doesn't allow this. A child who died should have been born.
We can't let it pass when the birthdate is missing.
There might be something else causing the bug.

cin_validator/rules/cin2022_23/rule_8535Q.py
```diff
@@ -31,12 +31,10 @@ def validate(
...
     # Return rows where DOB is prior to DOD
     condition1 = df[PersonBirthDate] > df[PersonDeathDate]
-    # Return rows with no DOB
-    condition2 = df[PersonBirthDate].isna()
 
     # df with all rows meeting the conditions
-    df_issues = df[condition1 | condition2].reset_index()
-
+    df_issues = df[condition1].reset_index()
+    print(df_issues)
```
tab1tha, line 37: to delete

tab1tha (changes requested):
The bug behaviour is worth a second look. 
The rule was already implemented as prescribed. I wonder why it doesn't work.

willLPD2I:
> Oh apologies, I have read it again. The first part of the rule doesn't allow this. A child who died should have been born. We can't let it pass when the birthdate is missing. There might be something else causing the bug.

@tab1tha Not true, not all pregnancies are carried fully to term. In these instances, the Expected DOB is after the date of death.

cin_validator/rules/cin2022_23/rule_8535Q.py
willLPD2I: Oh no, not the print 🤦

tab1tha:
> > Oh apologies, I have read it again. The first part of the rule doesn't allow this. A child who died should have been born. We can't let it pass when the birthdate is missing. There might be something else causing the bug.
> 
> @tab1tha Not true, not all pregnancies are carried fully to term. In these instances, the Expected DOB is after the date of death. Unless, as it's a query, we still want to flag this.

There is another variable for unborn babies, ExpectedPersonBirthDate. This rule only interacts with PersonBirthDate which is for children who have been born.
However, **if** you notice that the children who are wrongly flagged all have an ExpectedPersonBirthDate and no PersonBirthDate (i.e unborn), a filter can be added before the issue locations are returned. The filter should exclude any children in the list who have an ExpectedPersonBirthDate.

willLPD2I:
> > > Oh apologies, I have read it again. The first part of the rule doesn't allow this. A child who died should have been born. We can't let it pass when the birthdate is missing. There might be something else causing the bug.
> > 
> > 
> > @tab1tha Not true, not all pregnancies are carried fully to term. In these instances, the Expected DOB is after the date of death. Unless, as it's a query, we still want to flag this.
> 
> There is another variable for unborn babies, ExpectedPersonBirthDate. This rule only interacts with PersonBirthDate which is for children who have been born. However, **if** you notice that the children who are wrongly flagged all have an ExpectedPersonBirthDate and no PersonBirthDate (i.e unborn), a filter can be added before the issue locations are returned. The filter should exclude any children in the list who have an ExpectedPersonBirthDate.

They all have expected birth dates and no birth date meaning they died before they were born, so yes, I can add a filter for this and still keep the check for no birth date then

tab1tha (approved):

PR #360, opened 2023-02-23
Fix/key error 4001

tab1tha:


tab1tha:
Merging this in so that I can build the wheel file.

PR #361, opened 2023-03-01
mvp init file

tab1tha:
This adds an init file to the 23_24 folder so that the 23_24 rules can be implemented and run independently. 
The ability to import rules from 22_23 will be added in a later pull request.


tab1tha:
23_24 rules can be run using
`python -m cin_validator test --ruleset "rules.cin_2023_24`

willLPD2I (approved):

PR #363, opened 2023-03-02
ensure dates are passed as strings to fe

tab1tha:


tab1tha:
Fix bug where dates were appearing in exponential-like format.

PR #364, opened 2023-03-03
apply datetime_to_str recursively

tab1tha:
Converts all Timestamp objects (including those within tuples) to string dates so that they are human-readable in the user report and not epochs.


tab1tha:
This has been shown to work when tested in the frontend. The dates in the report are shown in the right format.
So, I'll merge it in.

PR #386, opened 2023-03-09
 bugfix 4014 to flag duplicate cin plans

### LINKED ISSUE #366: Bug: Rule 4014 - Incorrectly discounts duplicate CIN Plans
**Describe the bug**
Where there are two CIN Plans in CINplanDates that have the exact same `CINPlanStartDate` and `CINPlanEndDate` this is being filtered out by the rule coding and not flagged as an error.  This is different behaviour to the DFE CIN portal which does flag the duplicate CIN Plans as "overlapping dates".  The CIN Validator front-end also only appears to show the CIN Plan once despite existing twice in the source data.

**To Reproduce**
- Create multiple CINplanDates records for the same child with the same CINPlanStartDate and CINPlanEndDate.

**Expected behavior**
1. Both records in CINplanDates should be returned in the CIN Validator front-end.
2. Both records should be flagged as in-error due to overlapping dates.

**Proposed fix (Optional)**
Instead of excluding CIN Plans where `CINPlanStartDate` = `CINPlanStartDate` this should exclude plans where `ROW_ID` = `ROW_ID` (based on original table index).

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_4014.py#L64-L67

willLPD2I:
closes #366

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_4014.py
+++ b/cin_validator/rules/cin2022_23/rule_4014.py
@@ -61,7 +61,6 @@ def validate(
         suffixes=("_cinp", "_cinp2"),
     )
 
-
     # Use CINPlanStartDate to identify a CIN plan. Exclude rows where the ROW_ID is the same on both sides to prevent a plan from being compared with itself.
     df_merged = df_merged[df_merged["ROW_ID_cinp"] != df_merged["ROW_ID_cinp2"]]
 
@@ -90,7 +89,7 @@ def validate(
             df_merged["CINPlanStartDate_cinp2"],
         )
     )
-    
+
     # The merges were done on copies of cinp_df so that the column names in dataframes themselves aren't affected by the suffixes.
     # we can now map the suffixes columns to their corresponding source tables such that the failing ROW_IDs and ERROR_IDs exist per table.
     df_cinp_issues = (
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_4014.py
```diff
@@ -62,9 +62,7 @@ def validate(
     )
 
     # Use CINPlanStartDate to identify a CIN plan. Exclude rows where the CINPlanStartDate is the same on both sides to prevent a plan from being compared with itself.
```
SLornieCYC, line 64: Comment should be updated to reflect changed logic here.

cin_validator/rules/cin2022_23/rule_4014.py
```diff
@@ -166,6 +164,16 @@ def test_validate():
                 "CINPlanStartDate": "31/03/2001",
                 "CINPlanEndDate": pd.NA,
             },
+            {
```
SLornieCYC, line 167: This looks right to me!

SLornieCYC (approved):

PR #387, opened 2023-03-09
Bugfix 8816

### LINKED ISSUE #371: Bug: Rule 8816 - Incorrectly discounts duplicate CIN Episodes
**Describe the bug**
Where there are two CIN Episodes in CINDetails that have the exact same values for `CINreferralDate`, `CINclosureDate` (null) and `ReferralNFA` (0) this is being filtered out by the rule coding and not flagged as an error.  This is different behaviour to the DFE CIN portal which does flag the duplicate CIN Episode as "open but not latest".  The CIN Validator front-end also only appears to show the CIN Episode once despite existing twice in the source data.

**To Reproduce**
- Create multiple CINDetails records for the same child with the same `CINreferralDate`, `CINclosureDate` and `ReferralNFA`.

E.g. here a duplicate referral not NFA with no closure date:

```
<CINdetails>
<CINreferralDate>SAMEDATE</CINreferralDate>
<ReferralNFA>0</ReferralNFA>
</CINdetails>
<CINdetails>
<CINreferralDate>SAMEDATE</CINreferralDate>
<ReferralNFA>0</ReferralNFA>
</CINdetails>
```

**Expected behavior**
1. Both records in CINDetails should be returned in the CIN Validator front-end.
2. Both records should be flagged as in-error due to overlapping dates. (DFE CIN portal flags both but not the end of the world if CIN Validator only flags one - better than not flagging any!).

**Proposed fix (Optional)**
Instead of excluding CIN Episodes where `CINreferralDate` = latest `CINreferralDate` I think this should only exclude the Epsiode where `ROW_ID` = latest `ROW_ID` (based on original table index).

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8816.py#L51-L57

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8816.py#L67-L69

willLPD2I:
closes #371
Flags duplicate cin episodes. Only currently flags 1, where ideal is 2.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8816.py
+++ b/cin_validator/rules/cin2022_23/rule_8816.py
@@ -60,12 +60,12 @@ def validate(
     df_cin.reset_index(inplace=True)
 
     # filter the rows of interest
-    falseorzero = ["false", "0"]
+
     df_cin = df_cin[
-        (df_cin[CINclosureDate].isna()) & (df_cin[ReferralNFA].isin(falseorzero))
+        (df_cin[CINclosureDate].isna()) & (df_cin[ReferralNFA].isin(["false", "0"]))
     ]
     # check those that do not meet the requirements
-    condition = df_cin[CINreferralDate] != df_cin["latest_referral"]
+    condition = df_cin["ROW_ID"] != df_cin["latest_referral_ind"]
     df_cin = df_cin[condition]
 
     df_cin["ERROR_ID"] = tuple(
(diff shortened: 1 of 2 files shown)


cin_validator/rules/cin2022_23/rule_8816.py
```diff
@@ -146,19 +146,31 @@ def test_validate():
...
-                "LAchildID": "child4",
+                "LAchildID": "child4",  # 6, Fail, shared CINreferralDate
                 "CINclosureDate": pd.NA,
                 "CINreferralDate": pd.NA,
                 "CINdetailsID": "cinID3",
                 "ReferralNFA": "0",
             },
             {
-                "LAchildID": "child4",
+                "LAchildID": "child4",  # 7, Fail, shared CINreferralDate
```
SLornieCYC, line 156: Think this comment has been added to the wrong test case?

SLornieCYC:
I think only flagging one here is fair enough.  Not worth trying to re-write it to catch both.

willLPD2I:
You might notice I had some issues based on when  created each branch and solved branch conflicts in each with there being apparent changes to 4014 here, they should be fine now though! Serves me right for working on multiple branches at once.

PR #389, opened 2023-03-10
Bugfix/8839

### LINKED ISSUE #373: Bug: Rule 8839 - Not recognising ICPCnotRequired = 1
**Describe the bug**
Rule appears to not be filtering out Section47 records where `ICPCnotRequired` == 1.  I know there was a recent issue (https://github.com/data-to-insight/CIN-validator/issues/318) and fix for `ICPCnotRequired` == "true" but this is about the 1/0 data.

**Proposed fix (Optional)**
I think the df_check here needs to use **"1"** (string) rather than **1** (int) to correctly match the data passed from the xml source?

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8839.py#L47-L50

willLPD2I:
closes #373 
checks ICPCnotRequired == 1 as int and string

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8839.py
+++ b/cin_validator/rules/cin2022_23/rule_8839.py
@@ -46,7 +46,7 @@ def validate(
     # The rule originally asks for true, not 1, but an analyst coded it for 1, so their LA may use 1 and 0 instead, as such, it now check for either.
     df_check = df_check[
         df_check[ICPCnotRequired].isna()
-        | (~df_check[ICPCnotRequired].isin([1, "true"]))
+        | (~df_check[ICPCnotRequired].isin([1, "true", "1"]))
     ]
     # get all the locations where DateOfInitialCPC is null
     df_check = df_check[df_check[DateOfInitialCPC].isna()]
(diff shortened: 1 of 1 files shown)


cin_validator/rules/cin2022_23/rule_8839.py
```diff
@@ -46,7 +47,7 @@ def validate(
     # The rule originally asks for true, not 1, but an analyst coded it for 1, so their LA may use 1 and 0 instead, as such, it now check for either.
     df_check = df_check[
         df_check[ICPCnotRequired].isna()
-        | (~df_check[ICPCnotRequired].isin([1, "true"]))
+        | (~df_check[ICPCnotRequired].isin([1, "true", "1"]))
```
tab1tha, line 50: This does the trick.
I'll change the double check to use .astype(str) so that we don't have to manually type out multiple versions of the same element.
`(~df_check[ICPCnotRequired].astype(str).isin([ "true", "1"])`

tab1tha (changes requested):

tab1tha (approved):

Issue #385, opened 2023-03-08
Bug: Rule 8897 - Missing assessments with no factors


SLornieCYC:
**Describe the bug**
Rule is currently ignoring completed assessments where no factors have been recorded.

**To Reproduce**
Steps to reproduce the problem, e.g.:
- Create a completed Single Assessment record (i.e. with `AssessmentAuthorisationDate`) but no `AssessmentFactors`.
- This assessment should be flagged as an error (or query - per #343).

**Expected behavior**
CAssessments with no factors recorded should be flagged, **in addition to** as assessments that use factors that don't exist in the codeset (the latter is working).

**Proposed fix (Optional)**
The number of factors needs counting per assessment and a third condition adding that returns true where countoffactors == 0.

E.g. `condition1 & (~condition2 | condition3)`

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8897.py#L96-L104

willLPD2I:
I think this should be fixed by my PR #391 which builds assessments blocks for children with no assessment factors, I'd have fixed that one much more quickly if I'd looked at this too!

SLornieCYC:
Agree, this one looks to have been fixed by PR #391.

PR #391, opened 2023-03-13
bugfix 8670Q

### LINKED ISSUE #377: Bug: Rule 8670Q - Not returning errors
**Describe the bug**
Rule is returning no errors in CIN Validator whereas I get 74 errors in DFE portal from the same data.

**Proposed fix (Optional)**
Not sure what exactly is causing the issue however from looking at the python I have one initial query:

1. Does this actually do what it's suppposed to? The return from **england_working_days** is `pd.offsets.CustomBusinessDay(n=num_days, calendar=holiday_calendar)` but the function is being _subtracted_ from a date here rather than being applied to a date - so what is the function actually returning?

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8670Q.py#L50-L51

willLPD2I:
closes #377

8670Q wasn't returning errors, it looks like that was because it was calculating the business days not including the return date as one of the dates. It can be fixed, as in this PR, by just using 44 days in the function, but it would be far better to change the function to take the correct number of days, even if we did:

`return pd.offsets.CustomBusinessDay(n=num_days-1, calendar=holiday_calendar)`

or something silly like that, though I'm sure there's a tidier way.


willLPD2I:
Right, after a lot of testing, it turns out that, as well as the unrelated business days issue, the ingress didn't build blocks where there are no assessment_factors, I've got a temporary fix in here we can pretty up and dry up later when there's time that allows it to still make assessment blocks for children with no assessment factors. This is kinda funny as, if a kid has no completed assessment, they won't have assessment factors, so it allowed every single instance this rule was failing to be bypassed 😂

willLPD2I:
this should close #385 if the reason it wasn't picking up children with no assessment factors was, as would make sense, those children weren't being given assessment blocks.

cin_validator/ingress.py
```diff
@@ -330,14 +341,11 @@ def create_Assessments(self, cin_detail):
...
-                    }
-                    assessment_dict = get_values(elements, assessment_dict, assessment)
-                    # the get_values function will not find AssessmentFactors on that level so it'll assign it to NaN
-                    assessment_dict["AssessmentFactors"] = factor.text
-                    assessments_list.append(assessment_dict)
+                    assessment_block_maker(factor)
+            # else needed to build blocks in instances where assessments aren't completed which means there's no assessment factors to build with.
+            else:
+                for assessment in assessments:
+                    assessment_block_maker(assessment)
```
tab1tha, line 348: I'm here. Giving this some thought.
Since this else statement is within the a for-assessment loop already, it shouldn't have to loop through anything when assessment factors are not found and should instead behave similar to a create_section47 ingress process.

Putting a pin on it. Will come back after the release to refactor.

tab1tha (approved):

PR #392, opened 2023-03-13
bugfix 8675Q,

### LINKED ISSUE #378: Bug: Rule 8675Q - Not returning errors
**Describe the bug**
Rule is returning no errors in CIN Validator whereas I get 15 errors in DFE portal from the same data.

**Proposed fix (Optional)**
I think this line needs to accommodate "0" (string) as well as "false".

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8675Q.py#L49

willLPD2I:
closes #378 

 checks for 0 as well as false


tab1tha (approved):
It's nice that you've added the data that verifies the change. 
Thanks, Will.

tab1tha (approved):

PR #395, opened 2023-03-14
bugfix 2991Q 

### LINKED ISSUE #376: Bug: Rule 2991Q - [short rule description]
**Describe the bug**
Rule is returning no errors in CIN Validator whereas I get 58 errors in DFE portal from the same data.  

**Proposed fix (Optional)**
No idea what is causing my issue as described above but there are two things that jump out at me as queries from the rule coding (which are probably red herrings):

1. Should merged_df be a left join rather than an outer join?  i.e. all S47s and only matching assessments?

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_2991Q.py#L37-L44

2. Should df_ass_issues not be merging on `ROW_ID_ass` rather than `ROW_ID_S47`?

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_2991Q.py#L52-L57

willLPD2I:
closes #376 

Assessments was set to CINtable.CINdetails instead of assessments. Also, the merge used was outer instead of left. Before the fix,  the bug can could recreated by using CIN_census_2021 and removing the assessments block, an assessments block was still made.

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_2991Q.py
+++ b/cin_validator/rules/cin2022_23/rule_2991Q.py
@@ -5,7 +5,7 @@ import pandas as pd
 from cin_validator.rule_engine import CINTable, RuleContext, RuleType, rule_definition
 from cin_validator.test_engine import run_rule
 
-Assessments = CINTable.CINdetails
+Assessments = CINTable.Assessments
 Section47 = CINTable.Section47
 
 
@@ -34,11 +34,12 @@ def validate(
     df_ass.reset_index(inplace=True)
     df_47.reset_index(inplace=True)
 
+    #  If <Section47> module is present then <Assessment> module should be present.
     merged_df = df_47.merge(
         df_ass,
         on=[LAchildID, CINdetailsID],
         suffixes=["_47", "_ass"],
-        how="outer",
+        how="left",
         indicator=True,
     )
     merged_df = merged_df[merged_df["_merge"] == "left_only"]


cin_validator/rules/cin2022_23/rule_2991Q.py
```diff
@@ -5,7 +5,7 @@
 from cin_validator.rule_engine import CINTable, RuleContext, RuleType, rule_definition
 from cin_validator.test_engine import run_rule
 
-Assessments = CINTable.CINdetails
+Assessments = CINTable.Assessments
```
tab1tha, line 8: I agree.

cin_validator/rules/cin2022_23/rule_2991Q.py
```diff
@@ -34,11 +34,12 @@ def validate(
     df_ass.reset_index(inplace=True)
     df_47.reset_index(inplace=True)
 
+    #  If <Section47> module is present then <Assessment> module should be present.
     merged_df = df_47.merge(
         df_ass,
         on=[LAchildID, CINdetailsID],
         suffixes=["_47", "_ass"],
-        how="outer",
+        how="left",
```
tab1tha, line 42: It's okay to leave it as `outer` since the next line filters out the rows where the i`ndicator=="left_only"`. Hence, only the rows found in df_47 that are not found in df_ass will be left.
image

tab1tha (changes requested):
It's probable that the locations are not flagged because they are pushed to the CINdetailsID column which does not exist in the user interface. We created it for internal processes only. 

Pushing the error locations to the LAchildID column should fix this bug. The outer join works.

willLPD2I:
> It's probable that the locations are not flagged because they are pushed to the CINdetailsID column which does not exist in the user interface. We created it for internal processes only.
> 
> Pushing the error locations to the LAchildID column should fix this bug. The outer join works.

The errors came from the fact that assessments was brought in as CINdetails not assessments so it compared the wrong tables

cin_validator/rules/cin2022_23/rule_2991Q.py
SLornieCYC: Isn't it more efficient to remove the unnnecessary data at the earlier stage (merge) rather than pulling through irrelevant assessments just to discard them immediately afterwards?

willLPD2I:
> Isn't this join still wrong? (Can't flag in PR because not changed so below is current main)
> 
> https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_2991Q.py#L52-L57

@SLornieCYC I think it is wrong, but, and this is funny I suppose, I don't think it matters as, if the rule points out there's no assessment block, there's no assessment block to highlight even if the row ID is wrong?

cin_validator/rules/cin2022_23/rule_2991Q.py
tab1tha: You're right. Though the change in time/memory used will be quite small since we're working with small datasets, simply doing the left join is more efficient. 
Either way works. 
- outer join and filter by indicator or
- left join and remove indicator and filter line(s).
Whichever Will chooses is fine by me.

willLPD2I: My latest merge has gone back to outer so, rather than changing it again, lets leave it

tab1tha:
> > Isn't this join still wrong? (Can't flag in PR because not changed so below is current main)
> > https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_2991Q.py#L52-L57
> 
> @SLornieCYC I think it is wrong, but, and this is funny I suppose, I don't think it matters as, if the rule points out there's no assessment block, there's no assessment block to highlight even if the row ID is wrong?

That is, in fact, funny. It simply slipped through the cracks of the logic. 
Still, worth changing.

tab1tha:
Nice.
Thanks, Will.

I was just about to ask about your changes to fake data too. Then you reversed it. All good.

tab1tha (approved):

PR #415, opened 2023-03-21
Bugfix/8565

### LINKED ISSUE #369: Bug: Rule 8565 - False positives in CINEpisodes where DateOfInitialCPC is null
**Describe the bug**
Some CINEpisodes records are being incorrectly flagged against this rule for having a null `DateOfInitialCPC`, where the episode has a `CINclosureDate` recorded.  This is an optional field that is used for a specific subset of cases only (i.e. when a child transfers-in from another LA and was on CPP in that previous LA).

**To Reproduce**
- Have a child with a CINEpisode that has a `CINclosureDate` but no `DateOfInitialCPC`

**Expected behavior**
CINEpisode should not be flagged as in-error from the null `DateOfInitialCPC`.

**Proposed fix (Optional)**
I think condition8 here also needs to test `df["DateOfInitialCPC_CIN"].notna()`.

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8565.py#L193

### LINKED ISSUE #370: Bug: Rule 8565 - Section47 DateOfInitialCPC
**Describe the bug**
Rule does not check `DateOfInitialCPC` in Section47.

**To Reproduce**
- Have a child with a CINEpisode that has a `CINclosureDate` and a Section47 with a `DateOfInitialCPC` after the  `CINclosureDate`.

**Expected behavior**
Section47 should be flagged as in-error from the later dated `DateOfInitialCPC`.

**Proposed fix (Optional)**
I think condition1 here should be using `DateOfInitialCPC_S47` rather than `DateOfInitialCPC`.

https://github.com/data-to-insight/CIN-validator/blob/dc4559d2362e0efacfe47ebb9b4b3d538c7d0094/cin_validator/rules/cin2022_23/rule_8565.py#L186

willLPD2I:
closes #370 

8565 will now fail rows where DOICPC_S47 is after the CinClosureDate, where before it checked DOICPC_CIN twice, as it was merged as both DOICPC and COICPC_CIN (so with and without suffixes)

edit: should also closes #369

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8565.py
+++ b/cin_validator/rules/cin2022_23/rule_8565.py
@@ -58,7 +58,7 @@ CINPlanEndDate = CINplanDates.CINPlanEndDate
 @rule_definition(
     code=8565,
     module=CINTable.ChildProtectionPlans,
-    message="Activity shown after a case has been closed",
+    message="Activity shown after a case has been closed.",
     affected_fields=[
         CINclosureDate,
         DateOfInitialCPC,
@@ -181,7 +181,7 @@ def validate(
             ],
         )
     )
-
+    print(df)
     # Return those where dates don't align
     condition1 = df[CINclosureDate] < df[DateOfInitialCPC]
     condition2 = df[CINclosureDate] < df[AssessmentActualStartDate]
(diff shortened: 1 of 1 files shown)


willLPD2I:
should now close #369 

HOWEVER, due to the way that the errors are sent to the FE, the validator may still highlight incorrect fields when they have failed for a different field

tab1tha (changes requested):
These changes do fix the bugs for all intents and purposes.

Taking a step back, the merges and joins in this rule will be a source of mutiple edge cases. First, the tables do not need to be all joined up as they can be independently compared with the CINclosureDate. 
Secondly, for every merge, there is a possibility of losing data. Merges are inner-joins by default and we can't be sure that configuring all of them to be outer-joins is a better idea.

This deserves a re-write, we might as well do it now.

tab1tha (approved):

PR #425, opened 2023-04-05
ROW_ID to row_id

tab1tha:


tab1tha:
this has been changed in the final df because since "ROW_ID" is in string format, the code editor's "replace all occurences" hack doesn't successfully rename it throughout the repo.

Doing it one file at a time is prone to introducing bugs.

tab1tha:
I'll merge this in so that we can do a release.
No core functionality of the tool is affected.

PR #426, opened 2023-04-05
Refactor/remove ERROR_ID

tab1tha:
Location linking in the frontend is seen to work to a desirable level without ERROR_IDs and incorporating them has a high cost-benefit ratio so it is being de-prioritized.

For now, the ERROR_ID has been left to show up in the user report so that users can filter by ERROR_ID to levels of linking that are more specific than table-column-row. We'll need to find out if users use it in this way, else it can be removed from the user report too.

The concept of ERROR_IDs is tightly coupled with all ,but one, rule types. So totally deleting it from the backend is a significant piece of work.

Effect of change:
The command line run-all function returned 1699 rows before this change. Now, it returns 1690 rows as rows ERROR_ID is no longer used as a marker of uniqueness data entries with the same table-column-row are dropped as duplicates irrespective of their ERROR_IDs.

### DIFF
--- a/cin_validator/cin_validator_class.py
+++ b/cin_validator/cin_validator_class.py
@@ -247,6 +247,9 @@ class CinValidationSession:
         self.full_issue_df = include_issue_child(self.full_issue_df, raw_data)
         self.user_report = create_user_report(self.full_issue_df, raw_data)
 
+        # regularise column names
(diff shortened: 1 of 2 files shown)


tab1tha:
I'll merge this in so that we can do a release but feel free to flag it and we can talk extensively about what the changes mean.
No core functionality of the tool is affected.

PR #442, opened 2023-04-21
fix: error when no rule fails.

tab1tha:
1. https://github.com/data-to-insight/CIN-validator/pull/442/commits/d7657cbee74a914deacdffe81be38c153271cb6b initialises full_issue_df with all the column names it is supposed to have so that even if the structure of some rules doesn't generate a column, keyErrors wouldn't be raised when full_issue_df is processed. `python -m cin_validator run-all fake_data\fake_CIN_data.xml --select "100"` which returns a type 0 issue table (no ERROR_ID or LAchildID) now runs without errors.
2. https://github.com/data-to-insight/CIN-validator/pull/442/commits/6147036ff36009f5354cf3b33b22fcaef923fbaf does the same for full report such that `python -m cin_validator run-all fake_data\fake_CIN_data.xml --select "1520"`  or select 2883, la-level rule and rule-that-passes respectively, no longer break because the "values_flagged column cannot be found" when creating the user report.

### DIFF
--- a/cin_validator/cin_validator_class.py
+++ b/cin_validator/cin_validator_class.py
@@ -106,6 +106,11 @@ def include_issue_child(issue_df, cin_data):
         # work around: ensure that columns from both sources have the same type to prevent merge error
         table_df["ROW_ID"] = table_df["ROW_ID"].astype("int64")
         linker_df["ROW_ID"] = linker_df["ROW_ID"].astype("int64")
+
+        if not linker_df.empty:
+            # if failing locations have been found, remove the dummy LAchildID column to make way for the real one.
+            table_df.drop(columns="LAchildID", inplace=True)
+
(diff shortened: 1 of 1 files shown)


tab1tha:
type one rule. no LAchildID or ERROR_ID
`python -m cin_validator run-all fake_data\fake_CIN_data.xml --select "100"`

- [ ] full_issue_df generated
- [ ] user_report generated

La-level rule. No LAchildID or ERROR_ID. no values_flagged.
`python -m cin_validator run-all fake_data\fake_CIN_data.xml --select "100"`
- [ ] full_issue_df generated
- [ ] user_report generated

when no errors are found. No LAchildID or ERROR_ID. no values_flagged.
`python -m cin_validator run-all fake_data\fake_CIN_data.xml --select "8868"`

PR #452, opened 2023-06-08
Refactor/validation outputs

tab1tha:


tab1tha:
The changes in this pull request are part of syncing the structure of the CIN and LAC backends.
rpc_main was configured to return json but then that fails on Pyodide which doesn't accomodate NaNs so we reverted back to the string wrapping.

PR #457, opened 2023-08-31
fix: adding missing app.call decorator

dotloadmovie:


rpc_main.py
```diff
@@ -63,6 +63,7 @@ def generate_tables(cin_data: dict) -> dict[str, dict]:
     return cin_data_tables
 
 
+@app.call
```
tab1tha, line 66: it's the little things !

tab1tha (approved):

PR #459, opened 2023-09-07
Fix/stringify codes

tab1tha:
The previous change made on this branch, to change the rule codes to strings at the point where they are read into the registry, caused the in-rule-file tests to fail.
This was because, the type conversion happened in the validation func so that by the time the test_validation function was being run, the rule code had changed already.
For example, for a rule defined with `code=100` (int) the @rule_definition decorator will read it in, convert, then store it as a string. such that by the time we get to the `test_validate` function,
assert `rule_code == 100` will fail. instead we'd have to do assert `rule_code= "100"`

Considerations:
- The goal is to have rule codes all as strings by the time they reach the frontend. It is fairly straightforward to convert the rule_code column to str just before sending it over. However, isn't intuitive to a new developer who enters the codebase.
- If we want rule codes as strings then they should only be entered as strings. Not strings sometimes and ints other times. Consistency is a good thing to lean toward. 

The changes in this pull request implement the second consideration and updates the type expectation in the definition of `rule_definition` such that it expects only str rule codes and will flag if otherwise.

### DIFF
--- a/Documentation/debug_n_transform_scripts/stringify_rule_codes.py
+++ b/Documentation/debug_n_transform_scripts/stringify_rule_codes.py
@@ -1,0 +1,31 @@
+import json
+import re
+
+
+def extract_rule_code(filepath):
+    pattern = r"rule_(\d+)\.py"
+    match = re.search(pattern, filepath)
+    if match:
+        return match.group(1)
+    else:
+        return None
+
+
+with open("files_failed.json", "r") as f:
+    filepaths = json.load(f)
+# filepath = "C:\\Users\\tambe.tabitha\\CIN-validator\\cin_validator\\rules\\cin2022_23\\rule_8940.py"
+
+for filepath in filepaths:
+    print(f"+++++++++++++++++++++++++++++++{filepath}+++++++++++++++++++++++++++++++")
+    try:
+        rule_code = extract_rule_code(filepath)
+
+        with open(filepath, "r") as f:
+            file_text = f.read()
+            file_text = file_text.replace(f"{rule_code}", f"'{rule_code}'")
+        with open(filepath, "w") as f:
+            f.write(file_text)
+            print(f"updated {rule_code} to '{rule_code}'")
+    except Exception as e:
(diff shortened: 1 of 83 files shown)


tab1tha:
Again, how were these failing tests not caught when [the earlier pull request](https://github.com/data-to-insight/CIN-validator/pull/458) was made?
Somehow they managed to pass Github automated testing.

SLornieCYC:
> Again, how were these failing tests not caught when [the earlier pull request](https://github.com/data-to-insight/CIN-validator/pull/458) was made? Somehow they managed to pass Github automated testing.

Are the github actions configured to [checkout the relevant branch](https://github.com/actions/checkout#checkout-a-different-branch) when they run?  I think actions/checkout uses the repository main branch unless told otherwise.

tab1tha:
Thank you for digging into this @SLornieCYC 
The documentation you linked to is helpful. I've gone through, and it seems that if we implement it to point to the branch we're on, we'll have to edit the yaml file each time we create a new branch. 
Initially, it seemed that writing `my-branch` as is done in the tutorial will guess what branch you are on but when I searched the repo, I [couldn't find any reference to a my-branch variable](https://github.com/search?q=repo%3Aactions%2Fcheckout%20my-branch&type=code) apart from the one stated in the documentation. So it is a placeholder that has to be manually edited each time.

However, if we leave it to run on on the main branch, as it does now, then it checks if the incoming changes might cause the main branch to fail. This works sometimes (since some pull requests have failed tests while being on non-main branches) but seems to not work all the time.

SLornieCYC:
Can't you use the github [environment variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) there in place of `my-branch`?

E.g. something like: (Not 100% sure on the right syntax)

```
uses: actions/checkout@v4
with:
ref: ${{ github.head_ref || github.ref_name }}
```
https://stackoverflow.com/questions/60300169/how-to-get-branch-name-on-github-action
https://stackoverflow.com/a/71158878

PR #460, opened 2023-09-25
revise pandas version.

tab1tha:
- adjust to pandas version that doesn't break rule logic. (Above pandas 2.0, some rules begin to fail which might be due to a change in the behaviour of the pandas methods that they use.)
- remove explicit definition of Numpy version. Numpy isn't used in the tool and is only included as a dependency for pandas. So it is left to automatically adjust it's version to what the installed pandas version requires.


franciscobmacedo (commented):
looks good!

PR #464, opened 2023-09-28
issues_df without header issues

tab1tha:
Header issues were selected and included in a diffferent object - multichild_issues. However, they weren't removed from the general issues_df object. 
The changes in this pull request do that.

Also, when the changes were being tested on prpc, it was found that a few prpc dependencies had not been added to poetry so those have been added.

### DIFF
--- a/cin_validator/cin_validator.py
+++ b/cin_validator/cin_validator.py
@@ -284,10 +284,14 @@ class CinValidator:
         )
         self.full_issue_df.reset_index(drop=True, inplace=True)
 
-        # combine multichild issues
+        # remove header issues from main dataframe and transfer to no-child-id dataframe
         header_issues = self.full_issue_df[
             self.full_issue_df["tables_affected"] == "Header"
         ]
+        self.full_issue_df = self.full_issue_df[
+            self.full_issue_df["tables_affected"] != "Header"
+        ]
+        # combine multichild issues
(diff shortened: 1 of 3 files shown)


franciscobmacedo (approved):

cin_validator/cin_validator.py
```diff
@@ -284,10 +284,14 @@ def __init__(
         )
         self.full_issue_df.reset_index(drop=True, inplace=True)
 
-        # combine multichild issues
+        # remove header issues from main dataframe and transfer to no-child-id dataframe
         header_issues = self.full_issue_df[
```
cyramic, line 288: Is there a reason this isn't `self.header_issues` like the surrounding calculated values? Is there no chance this will ever be used outside of this function?

This is fine if that's the case. I just wanted to flag it just in case it was a mistake.

cyramic (approved):

cin_validator/cin_validator.py
tab1tha: Good shout.
This is the last step in the computation before values are returned from the backend. So there is no chance that another function after this will need `header.issues`.

tab1tha: Also, we don't need header issues on their own. We select them out of the main issues df to combine them with la issues which were processed separately. So the combined df, `self.multichild_issues`, is added as an attribute that can be accesses from outside the class.

PR #466, opened 2023-10-04
Workflow checkout branch

SLornieCYC:
Allow github actions to checkout the branch that it is being run on, instead of main. (Ref discussion in comments on end of #459).

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -7,6 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref || github.ref_name }}
       - uses: psf/black@stable
         with:
           version: "22.8.0"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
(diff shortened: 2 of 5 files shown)


SLornieCYC:
The linting action now works:

image

### File changed
image

### The failure
image

image

SLornieCYC:
The validator_tests action is more problematic.

image

1) Branch checkout is working fine, **however**:

2) The validator test defaults to rules in the cin2022_23 folder so won't pick up changes to rules in more recent years.  (See test 1 which didn't identify the assert).

3) The devcontainers task that runs `python -m cin_validator test` doesn't return a failure to the action when it finds an assert. (See test 2 which did identify the assert but still reported as a pass when the action completed).

image

SLornieCYC:
And the tests action has the same issue as validator_tests.  The job doesn't return a failure to the action when it finds an assert.

image

tab1tha:
Thank you for making the change @SLornieCYC and for keeping such a good log. 
It review easy. 

I have looked into the second case where the action doesn't fail as expected, though there is clearly a failing test. I do not yet know why that happens.

SLornieCYC:
Realised that this isn't actually necessary at all (having looked in more detail at the logs in the executed action on other PRs/commits I can see they all show it checking out the correct branch anyway).  Therefore closing this PR.  

There's _potentially_ still something to resolve around propogating errors up through the action from asserts and the test functions handling multiple years of rules but I need to look into that some more before creating new git issues.

Issue #469, opened 2023-10-23
Error code 2886Q [Amended 2024 v1.0]


SLornieCYC:
#### Error code
2886Q
#### Module
Return level validation
#### Message
Please check and either amend data or provide a reason: Percentage of children with no ~~gender~~ **sex** recorded is more than 2% (excluding unborns)
#### Validation check
The sum of: 

the number of child records where ~~`<GenderCurrent> (N00065)`~~ **`<Sex> (N00783)`**, is equal to **U or** missing, and the <ExpectedPersonBirthDate> (N00098) is equal to missing, divided by the total number of child records should be less than or equal to 0.02. Validation should be triggered at LA level.

For further info, refer to the DfE validation document [here](https://www.gov.uk/government/publications/children-in-need-census-2024-to-2025-specification)

PR #472, opened 2023-11-22
Adds Sex to ChildIdentifiers table in api and ingress, adds sample rule to test

### LINKED ISSUE #468: Gender vs Sex recording (from 2024/25)
From the 2024/25 return the existing `<GenderCurrent> (N00065)` field in the XML specification is replaced with `<Sex> (N00783)`.  

This will require changes to **ingress.py**, **rule_engine/__api.py** and possibly other files (anything on FE side?).  The existing `<GenderCurrent>` field needs to continue to exist and function for 2023/24 (and earlier? Not sure what tool policy is on accessibility of previous return years).

Affected validation rules will be covered in separate issues.

WillLP-code:
closes #468 #469 

Adds <Sex> (in addition to keeping <GenderCurrent>) to both the api and ingress.

This works because when building the tables, if a child block has no <GenderCurrent>, it'll leave it as a pd.NA, and the same for <Sex>. This is a much easier fix than checking the return year sent from the FE and changing the ingress based on this. 

Caveat: the way the rules list builds means that any rules which check to see <GenderCurrent> isn't empty carried over from previous years need to be updated and added to the 24/25 ruleset using <Sex> so that errors aren't thrown for them. This would be the case even if the ingress was done by checking the return year anyway, and this way, until that's all done, we get errors thrown by the validator, not thrown by Python when it can't find columns.

SO: every rule with <GenderCurrent> needs to be updated and added to 24/25 using <Sex> (and the new gender codes). This will be a separate PR once this is merged.

Also, adds 2886Q as a proof of concept and an updated XML so that the run command in the CLI can be used for 24/25.


AlistairHerbert (approved):
2886Q appears to refer to "gender" in text, when it's about "sex" in code - now fixed.

PR #480, opened 2024-05-21
Correct handling of unborns in 8535Q

### LINKED ISSUE #478: Bug: Rule 8535Q - does not flag unborns
**Describe the bug**
Rule does not trigger when client record has a DOD but no DOB.  Looks like this was changed in https://github.com/data-to-insight/csc-validator-be-cin/pull/357 however the previous logic was correct.  There is a specific bypass for this query in the DFE portal for unborns.

> 8535Q: All records have been checked and can confirm that the data has been amended (where required). All remaining queries relate to pregnancies that did not proceed to full term or the child died before or during birth.

**Expected behavior**
If client has a DOD and no DOB this should be flagged as in-error.

**Proposed fix**
Reinstate the original code from before https://github.com/data-to-insight/csc-validator-be-cin/pull/357.  Note that the commit/PR **cannot** be reverted in its entirety because there are appear to be other changes included.

SLornieCYC:
Closes #478

### DIFF
--- a/cin_validator/rules/cin2022_23/rule_8535Q.py
+++ b/cin_validator/rules/cin2022_23/rule_8535Q.py
@@ -7,7 +7,6 @@ from cin_validator.test_engine import run_rule
 
 ChildIdentifiers = CINTable.ChildIdentifiers
 PersonBirthDate = ChildIdentifiers.PersonBirthDate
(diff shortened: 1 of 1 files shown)


WillLP-code (approved):
The code looks good, and I'll trust you on the rule logic here!

PR #508, opened 2026-01-05
Pre proceedings ingress

WillLP-code:
Adds the code needing to ingress the pre proceedings block of the CIN xml before adding the new rules

Note: there were definite typos in the DfE's published guidance [https://assets.publishing.service.gov.uk/media/695274dff828b4d60df50624/2026_to_2027_CIN_Census_Specification_v1-1.pdf](here)  I'm not saying they explain every mistake I may have made, but they may explain some.


SLornieCYC (approved):
This all looks fine to me, comparing to existing logic for other modules.

